PR #18, opened 2022-08-19
NEO-1070 - Remove sort icon, only render arrow up/down icon when column is sorted

mrazian85:
[Link to updated components in Deploy Preview](https://deploy-preview-18--neo-react-library-storybook.netlify.app/?path=/story/components-table--default)

This PR is for the work in [NEO-1070](https://jira.forge.avaya.com/browse/NEO-1070). It removes the 'sort' icon from Table Headers. It also adds logic so that the up/down sorting arrow icons are only displayed when the column is explicitly sorted. It also adds tests to ensure this behaviour works as intended.

**EXAMPLE IMAGE(S):**
Screen Shot 2022-08-19 at 5 33 52 PM

Table Sorting Icon gif


enrique-prado (approved):
Code looks great but there is a warning in the build.

joe-s-avaya (approved):
Nice work! That was a much simplier fix than I thought it would be. Also, really appreciate you taking the time to make good tests. Use of `querySelector` is not ideal, but I'm not seeing any other way to test that off the top of my head, so yeah, looks great! 

Also, you can just add this line directly above the `any` issue if you want to fix that in this PR. Else I'll add the line in the next PR.
`// eslint-disable-next-line @typescript-eslint/no-explicit-any`

yangAtSpoken (approved):

mrazian85:
@matthewAvaya it looks like the best way to fix the width of the Table cells is by using CSS, which would require a change in the CSS library.

I suggest that, if there are more changes to the Table styling to be done, then this change can be added to them and captured in a separate ticket for that work.

matthewAvaya (approved):
Approved.

PR #19, opened 2022-08-22
update library rules after I did some reading on Vite

joe-s-avaya:
https://vitejs.dev/guide/build.html#library-mode


joe-s-avaya:
storybook build issues, closing for now

PR #36, opened 2022-09-16
cleanup stories, make some minor improvements

joe-s-avaya:

Removed a duplicate story and did some minor cleanup. Trying to keep these PRs nice and small.


src/components/TopNav/TopNav.tsx
```diff
@@ -77,6 +83,8 @@ export const TopNav = ({
       </div>
 
       <div className="neo-nav" style={{ alignItems: "center" }}>
+        {tabs}
```
joe-s-avaya, line 86: moving `tabs` to right side as, even though the DP says to have it on the left, we use it on the right

src/components/TopNav/TopNav_shim.css
```diff
@@ -0,0 +1,3 @@
+.neo-navbar .neo-nav .neo-tabs {
+  height: 57px;
+}
```
joe-s-avaya, line 3: tabs were covering the bottom border, so I'm making them one pixel smaller than what the top nav is hardcoded to

yangAtSpoken (approved):

src/components/TopNav/TopNav.tsx
mrazian85: It would be good to have a Story that showed when the TopNav has a combination of elements, as opposed to showing them one-by-one.

For example, a TopNav with logo, search, tabs, buttons and avatar

joe-s-avaya: Hmmmm... I'm not sure I agree due to that not being a realistic story... I'll play with it.

PR #44, opened 2022-09-29
NEO-1171 AppLayout component Part2

enrique-prado:

Footer Area has been added, if provided it will render at the bottom of the screen.
The only area (prop) that is required is mainContent. Any optional areas that are not provided will not be rendered. See storybook stories in Layout category.

image

**IMPORTANT NOTES:**
- I'm getting unexpected errors in my basic render tests. It will take time to investigate a fix so I will add tests in Part 3.

### DIFF
--- a/src/components/Layout/AppLayout.tsx
+++ b/src/components/Layout/AppLayout.tsx
@@ -51,10 +51,10 @@ export const AppLayout = ({
     >
       {(Areas) => (
         <>
-          <Areas.Topheader>{header}</Areas.Topheader>
-          <Areas.Leftpanel area={"leftpanel"}>{leftPanel}</Areas.Leftpanel>
+          {header && <Areas.Topheader>{header}</Areas.Topheader>}
+          {leftPanel && <Areas.Leftpanel area={"leftpanel"}>{leftPanel}</Areas.Leftpanel>}
           <Areas.Main area={"main"}>{mainContent}</Areas.Main>
-          <Areas.Rightpanel area={"rightpanel"}>{rightPanel}</Areas.Rightpanel>
+          {rightPanel && <Areas.Rightpanel area={"rightpanel"}>{rightPanel}</Areas.Rightpanel>}
           <Areas.Footer area={"footer"}>Footer goes here</Areas.Footer>
         </>
       )}
--- a/src/components/Layout/AppLayout.tsx
+++ b/src/components/Layout/AppLayout.tsx
@@ -3,6 +3,7 @@ import { ReactNode, useEffect } from "react";
 
 export interface AppLayoutProps {
   desktopAreas?: string;
+  footer?: ReactNode;
   header?: ReactNode;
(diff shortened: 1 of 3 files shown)


src/components/Layout/AppLayout.tsx
```diff
@@ -64,18 +60,19 @@ export const AppLayout = ({
       areas={mobileAreas || defaultMobileAreas}
       areasMd={desktopAreas || defaultDesktopAreas}
       areasLg={desktopAreas || defaultBigscreenAreas}
-      templateRows={"60px 1fr 30px"}
+      height={height}
+      templateRows={"auto 1fr auto"}
       templateCols={"1fr auto"}
-      templateColsMd={"250px 1fr auto"}
+      templateColsMd={"auto 1fr auto"}
```
mrazian85, line 66: Do we want/should we allow users to access & override these values?

mrazian85:
I like this a lot! Left a question/discussion point

src/components/Layout/AppLayout.tsx
enrique-prado: Good question. In a sense they could override the auto values by passing a leftPanel (or rightPanel) component with a fixed size... the "auto" will simply resize to the size of the component that was passed in.  The center column "1fr" will simply take whatever horizontal space is available.

enrique-prado: Hi Mo, I replied to your question in the PR. Let me know if that answers it or if you need more info. thanks.

mrazian85: Yes, thanks.

I am happy with this. I think that, other than needing some tests (which I assume will come later?), this Component in particular will require feedback from our users in order to better understand how they are using it, and for us to make the necessary improvements. So, the sooner we get this out in the wild the better.

mrazian85 (approved):

src/components/Layout/AppLayout.tsx
enrique-prado: Yes, customer feedback will be key... and yes, the test will come later once I figure out a workaround for the failures.

src/components/Layout/AppLayout.tsx
```diff
@@ -64,18 +60,19 @@ export const AppLayout = ({
       areas={mobileAreas || defaultMobileAreas}
       areasMd={desktopAreas || defaultDesktopAreas}
       areasLg={desktopAreas || defaultBigscreenAreas}
-      templateRows={"60px 1fr 30px"}
+      height={height}
+      templateRows={"auto 1fr auto"}
```
joe-s-avaya, line 64: nit: using `{"ish"}` is doing two calculations, strongly recommend updating all `{"` to just `"`
```suggestion
      templateRows="auto 1fr auto"
```

enrique-prado: Ah, good catch! done.

PR #60, opened 2022-11-02
rework `Chip` and `ChipContainer`

joe-s-avaya:
[Link to updated Chip Stories](https://deploy-preview-60--neo-react-library-storybook.netlify.app/?path=/story/components-chips)

Nearly all files have been renamed, and the only one that wasn't that needs to be paid close attention to is [Chips.tsx](https://github.com/avaya-dux/neo-react-library/blob/b95264c99b7422d1caf0de54c532981a06e1bea9/src/components/Chip/Chip.tsx), which I recommend not looking at the diff for, but looking at directly.

Unit test code coverage is 100% for the Chip and Chip Container.

Chip's are no longer buttons as that breaks a11y. But they do accept a `onClose` prop for when the user makes the chip closable (the close icon is a button, which also has an aria label).

Note that I have _not_ done the "[Expandable Chip](https://design.avayacloud.com/components/web/chip-web#ExpandableChips)" as I checked with Matt and he is unsure if we need to support that option. If Matt comes back and tells me that we do need it I'll add it.

**GIF Example**:


matthewAvaya:
Default
- Add 8px padding between each chip example (for all examples)

Closable
- Do we need a chip for 'Not Closable'? Seems odd that it's under the Closable category. 

Chips Container Examples
- Avatar looks big. On the design site, the avatar is scaled at 28x28px.
- I believe the title for the section with the avatars should be 'Chips with Avatar' like it is on the design portal. 

Templated
- What's the difference between this chip compared to the chip under the Default category? They look identical. 

Has the accessibility part of this been done? The screen reader doesn't work and the focus order seems incomplete.

yangAtSpoken (commented):
The changes lost some of the nuances:
* icon chip supports rtl; closable does not;
* when chips are inside chip container, they need to have neo-chips__item class.
* avatar needs to have size=sm attribute.
* Tooltip is gone.

joe-s-avaya:
> Default
> 
> * Add 8px padding between each chip example (for all examples)
> 
> Closable
> 
> * Do we need a chip for 'Not Closable'? Seems odd that it's under the Closable category.
> 
> Chips Container Examples
> 
> * Avatar looks big. On the design site, the avatar is scaled at 28x28px.
> * I believe the title for the section with the avatars should be 'Chips with Avatar' like it is on the design portal.
> 
> Templated
> 
> * What's the difference between this chip compared to the chip under the Default category? They look identical.
> 
> Has the accessibility part of this been done? The screen reader doesn't work and the focus order seems incomplete.


Will add padding between chips when in a container.

In the "Closable" section, I added the "Not Closable" chip just so that the user can easily see the difference between the two. Especially when tabbing. 

Avatar is fixed. Title of that section is also fixed.

Templated is there for playing with the different options for a chip. I've added a GIF example in the description.

A11y has been done. I'll ping you about what exactly you're seeing.

joe-s-avaya:
> The changes lost some of the nuances:
> 
> * icon chip supports rtl; closable does not;
> * when chips are inside chip container, they need to have neo-chips__item class.
> * avatar needs to have size=sm attribute.
> * Tooltip is gone.

RTL: I see no reason why that would be. So I'm reluctant to make that change unless I can find a compelling reason to add it.
`neo-chips__item`: will add, I missed that.
Avatar size: oh yeah, good call, updated.
Tooltip: there's no reason to have props for tooltip. The dev can just wrap the chip (or it's content) with a tooltip.

yangAtSpoken:
> > The changes lost some of the nuances:
> > 
> > * icon chip supports rtl; closable does not;
> > * when chips are inside chip container, they need to have neo-chips__item class.
> > * avatar needs to have size=sm attribute.
> > * Tooltip is gone.
> 
> RTL: I see no reason why that would be. So I'm reluctant to make that change unless I can find a compelling reason to add it. `neo-chips__item`: will add, I missed that. Avatar size: oh yeah, good call, updated. Tooltip: there's no reason to have props for tooltip. The dev can just wrap the chip (or it's content) with a tooltip.

RTL requirements are from https://design.avayacloud.com/components/web/chip-web 
icon can be placed on either left or right;
closeable chip always have x on the right;
@matthewAvaya can you confirm this?

src/components/Chip/Chip.tsx
```diff
@@ -1,33 +1,70 @@
...
   closable?: boolean;
-  id?: string;
-  onClick?: React.MouseEventHandler;
-  variant?: Variants;
-};
+  closeButtonAriaLabel?: string;
+  disabled?: boolean;
+  icon?: IconNamesType;
+  onClose?: MouseEventHandler<HTMLButtonElement>;
+  variant?: "alert" | "default" | "info" | "success" | "warning";
```
enrique-prado, line 16: I thought Matt said chips don't need variants at Wednesday's meeting. Has this changed?

yangAtSpoken:
Another requirement is that Avatar chip has does not allow icon.

src/components/Chip/ChipsContainer.tsx
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @see https://design.avayacloud.com/components/web/chip-web
+ * @see https://neo-react-library-storybook.netlify.app/?path=/story/components-chips
+ */
+export const ChipsContainer = ({ children }: ChipsContainerProps) => (
+  <div className="neo-chips">
+    {Children.map(children, (child, index) =>
+      cloneElement(child, {
+        key: index,
+        className: "neo-chips__item",
```
yangAtSpoken, line 26: Does this add to "className" or replace it?

src/components/Chip/Chip.tsx
```diff
@@ -1,33 +1,70 @@
...
+export interface ChipProps
+  extends Exclude<HTMLAttributes<HTMLDivElement>, "onClick"> {
+  avatarInitials?: string;
   children: string;
   closable?: boolean;
-  id?: string;
-  onClick?: React.MouseEventHandler;
-  variant?: Variants;
-};
+  closeButtonAriaLabel?: string;
```
yangAtSpoken, line 12: I do not think we have closable avatar chip per https://design.avayacloud.com/components/web/chip-web

src/components/Chip/index.ts
```diff
@@ -1,13 +1,2 @@
...
-export { ExpandableChip } from "./ExpandableChip";
-export { IconChip } from "./IconChip";
-export { ChipContainer } from "./ChipContainer";
-export { AvatarChip } from "./AvatarChip";
-export type { AvatarChipProps } from "./AvatarChip";
-export type { BasicChipProps } from "./BasicChip";
-export type { ClosableChipProps } from "./ClosableChip";
-export type { ExpandableChipProps } from "./ExpandableChip";
-export type { IconChipProps } from "./IconChip";
-export type { ChipContainerProps } from "./ChipContainer";
```
enrique-prado, line 13: Did all these files get renamed or deleted? I was expecting the former.

enrique-prado (commented):
Overall the code looks good but I'm a little hesitant to approve because of all the deleted files. It's unclear to me what happened to all that code, I was expecting these files to be renamed not deleted. If you can clarify that would be great. thanks.

src/components/Chip/Chip.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+export const Templated = Template.bind({});
+Templated.args = {
+  avatarInitials: "",
+  children: "Templated Example",
+  closable: false,
+  closeButtonAriaLabel: "Close",
+  disabled: false,
+  icon: undefined,
+  variant: "default",
+};
```
enrique-prado, line 94: I'm noticing we don't have a story where a chip is created dynamically like it typically happens in a real app.... basically you type a word in an input control and after you hit enter it creates a chip.  It would be nice to have one like that.

src/components/Chip/Chip.tsx
joe-s-avaya: That was the dropdown/expandable chip, which he is still gathering information about. So yes, we _do_ want variants still.

src/components/Chip/ChipsContainer.tsx
joe-s-avaya: Whoops, replaces it. Will update, good catch, thanks!

src/components/Chip/Chip.tsx
joe-s-avaya: Just checked with Matt, you are correct. Updated.

src/components/Chip/index.ts
joe-s-avaya: Deleted. Whereas before we have a separate file for every variant, now we have one single file that takes props. Thus I was able to reduce from many chip components to just two (`Chip` and `ChipContainer`).

src/components/Chip/Chip.stories.tsx
joe-s-avaya: Those exist in the [Select Creatable](https://deploy-preview-60--neo-react-library-storybook.netlify.app/?path=/story/components-select--creatable) stories. I don't feel that it's necessary to add a story for that, but if you think that it would add a lot to the DX, I'm happy to add it.

joe-s-avaya:
> > > The changes lost some of the nuances:
> > > 
> > > * icon chip supports rtl; closable does not;
> > > * when chips are inside chip container, they need to have neo-chips__item class.
> > > * avatar needs to have size=sm attribute.
> > > * Tooltip is gone.
> > 
> > 
> > RTL: I see no reason why that would be. So I'm reluctant to make that change unless I can find a compelling reason to add it. `neo-chips__item`: will add, I missed that. Avatar size: oh yeah, good call, updated. Tooltip: there's no reason to have props for tooltip. The dev can just wrap the chip (or it's content) with a tooltip.
> 
> RTL requirements are from https://design.avayacloud.com/components/web/chip-web icon can be placed on either left or right; closeable chip always have x on the right; @matthewAvaya can you confirm this?

@yangAtSpoken, just discussed with @matthewAvaya. The RTL ish is something that he is going to discuss with the Localization team and then get back to me on. For now he is comfortable leaving as is.

src/components/Chip/ChipsContainer.tsx
joe-s-avaya: Updated. I also noticed that I was extending the HTML Div Element, but I wasn't actually applying any of those props. Thus, I've also added `{...rest}` to both `Chip` and `ChipsContainer`.

src/components/Chip/Chip.tsx
enrique-prado: ok, sounds good

src/components/Chip/index.ts
enrique-prado: oh, I see. Nice!

src/components/Chip/Chip.stories.tsx
enrique-prado: cool. thanks.

enrique-prado (approved):

joe-s-avaya:
@matthewAvaya, I've made the updates we discussed, please reviews again: [Link to Chip stories](https://deploy-preview-60--neo-react-library-storybook.netlify.app/?path=/story/components-chips)

matthewAvaya (approved):

PR #70, opened 2022-11-08
Update Checkbox functionality to be more DX friendly 

joe-s-avaya:
[Link to updated Checkbox Group](https://deploy-preview-70--neo-react-library-storybook.netlify.app/?path=/story/components-checkbox-group)

Updated Checkbox to

- allow children (instead of `label` prop with same type)
- brought code coverage up to 100%

Updated Checkbox Group to

- allow just `label` or `groupName` (becomes the HTML `for` and `name` props) (this fixes the issue I'm having in the DPv3)
- properly use `aria-labelledby`
- brought code coverage up to 100%

**EXAMPLE IMAGE(S):**
Screen Shot 2022-11-08 at 12 28 23 PM

**IMPROVEMENT NOTES**:
The stories and tests could be improved in readability/maintainability. They aren't bad, but I think that there are more than necessary and the generated "docs" code isn't great. If you think that it is worth the time to update these, please note that in your review and I'll make another PR to clean it up.


src/components/Checkbox/Checkbox.tsx
```diff
@@ -96,7 +97,7 @@ export const Checkbox = forwardRef(
           id={id}
           checked={state === "mixed" || state || false}
           aria-checked={state || "false"}
-          aria-label={ariaLabel || `${label}`} // specify this and voiceover will announce state change
+          aria-label={ariaLabel || (children as string)}
```
yangAtSpoken, line 100: Casting children as string may not work well.  Maybe a 3rd party tool like https://github.com/sunknudsen/react-node-to-string
could help.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -107,11 +108,8 @@ export const Checkbox = forwardRef(
           {...rest}
         />
 
-        {/*
-        BUG: the Neo styles are all on the `label` element, so if there isn't a `label`,
-        there's no checkbox. Which is bad.
-      */}
-        <label htmlFor={id}>{label}</label>
+        {/* NOTE: the Neo styles are all on the `label` element, so if there isn't a `label` tag, there's no checkbox. Which is bad. */}
+        <label htmlFor={id}>{children}</label>
```
yangAtSpoken, line 112: This could be a good question for Terri today.  But the answer in this page says not many things could be children of <label>, https://stackoverflow.com/questions/18609554/is-div-inside-label-block-correct

src/components/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -1,47 +1,96 @@
...
+  if (!groupName && !label) {
+    handleAccessbilityError(
+      "CheckboxGroup: You must provide either a `groupName` or a `label` prop."
+    );
+  }
+
+  const htmlForName = useMemo(
+    () =>
+      label && !groupName ? label.toLowerCase().replace(" ", "-") : groupName,
+    [groupName]
```
yangAtSpoken, line 86: missing label

src/components/Checkbox/Checkbox.tsx
joe-s-avaya: I think you meant to write that the `<label>` tag should only contain a small subset of other HTML elements. Which isn't really a Terri question as she isn't a code person, but you could ask her, I just don't think that she'd know the answer.

But either way, I didn't change anything about what is passed into the `<label>` tag. It was originally `string | JSX.Element`, which is equivalent to `ReactNode`. So nothing has changed here, and if you want me to restrict the possible passed elements to a specific set, I think that should be a different PR. I'm also hesitant to do that as I'm not seeing a well written explanation of _exactly_ what children are and are not allowed. 
(from looking at [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) and the [html-spec](https://html.spec.whatwg.org/multipage/forms.html#the-label-element))

src/components/CheckboxGroup/CheckboxGroup.tsx
joe-s-avaya: I don't understand what you're saying here. If `label` exists, and there is no `groupName`, I use a modified label. Else, I know from the typings (and a11y `if` guard) that the `groupName` _must_ exist, and I use that.
Are you saying that there is a bug in my logic? I have tests for this, so it _should_ all be correct.

src/components/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -1,47 +1,96 @@
...
 }: CheckboxGroupProps) => {
+  if (!groupName && !label) {
+    handleAccessbilityError(
+      "CheckboxGroup: You must provide either a `groupName` or a `label` prop."
+    );
+  }
+
+  const htmlForName = useMemo(
+    () =>
+      label && !groupName ? label.toLowerCase().replace(" ", "-") : groupName,
```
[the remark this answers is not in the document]
joe-s-avaya: Oh, that's a great simplification. Will make the update.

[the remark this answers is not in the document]
joe-s-avaya: updated

src/components/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -1,47 +1,95 @@
...
   inline?: boolean;
   helperText?: string;
   error?: boolean;
   required?: boolean;
   onChange?: (e: React.ChangeEvent<HTMLInputElement>) => void;
 }
 
+type EnforcedAccessibleLabel =
+  | {
+      label: string;
```
yangAtSpoken, line 22: A label is usually associated with one input control.  The label in CheckboxGroup is not associated with any checkboxes inside this group.  It just displays group name.  I would suggest a boolean property, showGroupName (default to true), instead.

src/components/CheckboxGroup/CheckboxGroup.tsx
yangAtSpoken: It is not about handleAccessbilityError checking.  It is that general rule, i.e. whatever vars used inside a hook need to be in the dependency list.

src/components/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -1,47 +1,95 @@
...
+      groupName?: string;
```
yangAtSpoken, line 23: groupName should be a required field for a checkbox group.

yangAtSpoken (changes requested):
To hide group name, use a boolean instead as label is not an accessibility requirement for a checkbox group.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -20,17 +20,17 @@ interface BaseCheckboxProps
...
     }
   | {
-      label?: string | JSX.Element;
+      children?: ReactNode;
       "aria-label": string;
       "aria-labelledby"?: string;
     }
   | {
-      label?: string | JSX.Element;
+      children?: ReactNode;
```
enrique-prado, line 33: If you are casting children as string down in line 100, why not make children type 'string' instead?

enrique-prado (commented):
Just the one question about the children type.

src/components/CheckboxGroup/CheckboxGroup.tsx
joe-s-avaya: Yes, that makes sense, updated. I didn't understand what you were asking for.

src/components/CheckboxGroup/CheckboxGroup.tsx
joe-s-avaya: No for two reasons. 

One, a "label" should be a label and nothing more. Previously, we were inappropriately using the `groupName` as the `for`, "label", and `name`. Now, `label` is just the label of the group.*
Two, the `groupName` is the `name` prop of every checkbox in the group, and the `for` used on the `label`.

Thus, we should not use a boolean because the `for`+`name` need to be specific and follow best practices (ie, no spaces), and we need to allow the passing of one or both of `label` and `groupName`.

* unless no `groupName` is passed, then we modify the `label` to follow best practices and use that generated value for the `for`+`name` props. But even under that circumstance, the "label" is still just a label. We could just as easily use a generated value for the same, but this approach makes the DX slightly easier if you're looking through the DOM code.

joe-s-avaya: Also, the `label` in this case is associated with one "group" of inputs, as I believe is the best practice for both groups of checkboxes and groups of radio buttons.

src/components/CheckboxGroup/CheckboxGroup.tsx
joe-s-avaya: Why? We can easily generate it, there's no reason to force the user to pass it that I can think of.

src/components/Checkbox/Checkbox.tsx
joe-s-avaya: Fair question. I'm not sure if that passing `children` to the aria-label is correct, but there's a test for it so I don't want to undo that functionality as I presume that the original implementor knew what they were doing. Also, I don't want to force the `children` to be just a `string` as that would also remove previous functionality. 

However, you make a good point. So I've updated line 100 to only use `children` if it is a string type.

src/components/Checkbox/Checkbox.tsx
joe-s-avaya: updated

src/components/Checkbox/Checkbox.tsx
enrique-prado: ok, sounds good to me.

enrique-prado (approved):

src/components/CheckboxGroup/CheckboxGroup.tsx
yangAtSpoken: No, for + name is in correct. for + id is the correct combination.

src/components/CheckboxGroup/CheckboxGroup.tsx
yangAtSpoken: groupName is submitted to backend in groupName=value1+value2 form.  It should not be generated.

src/components/CheckboxGroup/CheckboxGroup.tsx
joe-s-avaya: @yangAtSpoken, the generated name is `htmlForName` on line 83. It is then used on lines 108, the `Checkbox` `name` fields, and line 116 the `htmlFor`/`for` `label` prop. Where do you see it being used as an `id` prop?

src/components/CheckboxGroup/CheckboxGroup.tsx
joe-s-avaya: @yangAtSpoken yes, but only if the user uses the default form and submit, which very rarely happens in my experience. And they are allowed to pass the group name, it's just not required. Thus, because people rarely use the default form submit, it shouldn't be required.

yangAtSpoken: I do not agree here. I think submit data correctly is the most important thing for a form control.  Thus do not the user guess what the group name is from a label.  And the label you added is not an accessibility requirement for this control.

src/components/CheckboxGroup/CheckboxGroup.tsx
yangAtSpoken: I am just saying htmlForName is not a good name. The better name here is groupName, which is already passed in.  Besides htmlFor is usually associate with a control's id.  So htmlForName seems a misnomer to me.

src/components/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -52,35 +100,32 @@ export const CheckboxGroup = ({
...
 
   return (
-    <NeoInputWrapper
-      data-testid="CheckboxGroup-root"
-      required={required}
-      error={error}
-    >
-      <label htmlFor={groupName}>{groupName}</label>
+    <NeoInputWrapper required={required} error={error}>
+      {label && <label htmlFor={htmlForName}>{label}</label>}
```
yangAtSpoken, line 116: htmlFor bounds to an element id, [label for attribute](https://www.w3schools.com/tags/att_label_for.asp)

src/components/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -52,35 +100,32 @@ export const CheckboxGroup = ({
...
-        />
-      );
-    });
-  };
+  const computeCheckboxesJsx = () =>
+    Children.map(children, (child, index) => (
+      <Checkbox
+        aria-describedby={helperTextId}
+        key={index}
+        name={htmlForName}
```
yangAtSpoken, line 108: htmlForName will end up on the "name" attribute of the checkbox.  It should end in id attribute for accessibility.  But checkbox already has a label.

src/components/CheckboxGroup/CheckboxGroup.tsx
joe-s-avaya: Per our discussion, will deal with in another PR as this will require a larger update

joe-s-avaya:
@yangAtSpoken, I'll deal with the a11y `label` for the group in another PR (per our discussion), and I think that I've dealt with all of the other issues you've seen. Please take another look and let me know if there's anything else you want me to update in this PR or the next one.

PR #83, opened 2022-11-30
Neo-1250 Part 1 tabs a11y update

yangAtSpoken:
[Tabs](https://deploy-preview-83--neo-react-library-storybook.netlify.app/?path=/story/components-tab--uncontrolled-active-tab-story)

Aria Rules to comply:
* tablist role can only have tab role inside;
* tabpanel must have one associated tab;
* no nested interactive roles: button role can not be child of a tab role;
* ul can not have tablist role;
* li must be inside of ul or ol;

Code changes:
* moved tablist role to tab nav div; thus moved carousel buttons out of tablist;
* changed ul and li to div;
* tab link changed to tab;
* close button is now just a span and an id is auto generated for it;

Unit Tests done locally with  "jest-axe": "7.0.0";
* all tab tests are passing;
* 8 tests failing due to Select component that should be fixed with next PR

Unit Tests done with  "jest-axe": "6.0.1";
* all tests are passing

Visual Tests:
* I do not see any visual difference by looking at storybook examples
* @matthewAvaya please see if there are any visual changes

jest-axe version is changed back to "6.0.1" so that we could merge this PR if we choose to.

As we discussed, close button is not a button any more.  Its mouse click is handled in parent tab.  Press 'backspace' or "X"  key will close the tab.  @matthewAvaya Please verify on this.


joe-s-avaya:
"[Disable|Enable Tabs](https://deploy-preview-83--neo-react-library-storybook.netlify.app/?path=/story/components-tab--controlled-active-tab-story)" story shows tab 2 selected, but shows content of tab 1.


All other stories are working properly (as far as I can tell), so I suspect that this is a minor issue in the story itself, not an issue with Tabs themselves. But it should be fixed ASAP.

But does not necessarily need to be fixed in this PR, just needs to be fixed; as it was not brought on by this change ([link to current neo-react storybook](https://neo-react-library-storybook.netlify.app/?path=/story/components-tab--controlled-active-tab-story) showing the same issue).

src/components/Tab/InternalTab.tsx
```diff
@@ -156,7 +157,8 @@ export const InternalTab = ({
 
       {closable && (
         <span
-          role="button"
+          role="tab"
```
joe-s-avaya, line 160: Having the close btn (even though it's not actually a `<button>`) be a tab feels very wrong as the screen reader will announce the number of tabs incorrectly; in the case of the "Basic Tabs" story, it announces _eight_ tabs, where it should only announce 6.

Does it throw a11y errors if you simply remove the `role` attribute?

src/components/Tab/InternalTabTypes.tsx
```diff
@@ -4,14 +4,15 @@ import { IconNamesType } from "utils";
...
-export interface InternalTabProps extends React.HTMLAttributes<HTMLLIElement> {
+export interface InternalTabProps extends React.HTMLAttributes<HTMLDivElement> {
   id: string;
   name: ReactNode;
   disabled: boolean;
   content: TabPanelProps;
   href?: string;
   icon?: IconNamesType;
   closable?: boolean;
+  closableId?: string;
```
joe-s-avaya, line 15: Why do we need this prop? I'm seeing it get assigned, but it doesn't look like it's being used for anything.

src/components/Tab/InternalTab.tsx
joe-s-avaya: It does throw an a11y error without a `role`. Dang, this makes no sense to me...uuuuurgh.

joe-s-avaya: Yeah, I'm playing with it and I'm not seeing any work-around for this. We'll have to get feedback from Terri on how to deal with this. In the mean time, please put a `// HACK: ish` comment here saying that we're waiting for feedback on how to deal with this.

I suspect that this may be a bug in `jest-axe`.

joe-s-avaya (approved):

src/components/Tab/InternalTabTypes.tsx
yangAtSpoken: Since closable button is a tab now, tablist needs to know the id of this tab.  Please see line 123 in Tabs.tsx and line 28 in TabUtils.tsx.

src/components/Tab/InternalTab.tsx
joe-s-avaya: [I've opened an issue in `jest-axe` asking for guidance](https://github.com/NickColley/jest-axe/issues/260).

src/components/Tab/InternalTabTypes.tsx
joe-s-avaya: ah, I see... ugh, this can't possible be right, but there's no other way around it... yuck... thanks for the explanation, make sense now.

matthewAvaya:
@yangAtSpoken Visually, this looks fine.

src/components/Tab/EventHandlers/KeyboardEventHandlers.ts
```diff
@@ -170,6 +146,18 @@ export const handleKeyDownEvent = (
         focus(ref, tabs[activeTabIndex].id);
       }
       break;
+    case Keys.BACKSPACE:
+    case Keys.X:
+    case Keys.x:
```
joe-s-avaya, line 151: This is so good, excellent idea and execution!

joe-s-avaya (approved):
This is an excellent idea! And executed beautifully as you updated all of the necessary related components (such as the `keys.ts`) and even reduced the code complexity in several areas! And also ensured to update all tests so that this is all self documenting. And I tested with VoiceOver and it all announces perfectly (and you allow the passing of `aria-label` but have an excellent default string). 

Really, really well done 💯

matthewAvaya:
The screen reader reads well, but how do you close the tab using the keyboard?

yangAtSpoken:
> The screen reader reads well, but how do you close the tab using the keyboard?

@matthewAvaya When the tab is on focus, press backspace or X.

matthewAvaya (approved):

src/components/Tab/InternalTab.tsx
```diff
@@ -123,12 +105,18 @@ export const InternalTab = ({
     }
   }, [focus, active, disabled]);
 
+  const defaultAriaLabel = closable
+    ? `You can press backspace or X to close tab ${name}`
```
mrazian85, line 109: 👌

mrazian85 (approved):
Nice! 👍

PR #84, opened 2022-11-30
NEO-1212 // add "esc" keyboard event to Tooltip component 

joe-s-avaya:
[Link to updated Tooltip stories](https://deploy-preview-84--neo-react-library-storybook.netlify.app/?path=/story/components-tooltip)

From some investigation of on NEO-1212, I realized that we need to support the "ESC" key usage to dismiss the tooltip. We can only support this in the react library, so I'm adding it here. Please note the comments I've made in `Tooltip.tsx` as I'm not 100% certain that I've created the "ideal" solution. 

Showing that ESC key works for focusable and non-focusable items, and with keyboard focus and mouse focus.


src/components/Tooltip/Tooltip.tsx
```diff
@@ -74,18 +77,56 @@ export const Tooltip = ({
...
+    [setAllowTooltip]
+  );
+
+  useEffect(() => {
+    document.addEventListener("keyup", onKeyUp, false);
+
+    return () => {
+      document.removeEventListener("keyup", onKeyUp, false);
+    };
+  }, []);
```
joe-s-avaya, line 106: I am unsure if this is the best way to ensure that we capture the keyboard event, but I tried several other options and this is the only one I could find that works. I worry that if you have multiple tooltip instances (such as the ["auto position" story](https://deploy-preview-84--neo-react-library-storybook.netlify.app/?path=/story/components-tooltip--auto-position)), it's flooding the document with "keyup" even watchers. Which feels sloppy to me.

However, I can't find any other way to ensure that the keyboard event is captured. If I put the "keyup" watcher anywhere _but_ the document, it won't get caught unless the user is keyboard focused on the tooltip child, which won't always be keyboard focusable (such as if the ["Templated" story](https://deploy-preview-84--neo-react-library-storybook.netlify.app/?path=/story/components-tooltip--templated)). 

Thus, this is the only solution I could find. Please let me know if you have any suggestions/ideas.

matthewAvaya:
I don't believe tooltips have a clickable action as they are typically displayed on hover. There have been discussions on a variant where the tooltip would stay open as there are clickable actions in the tooltip copy, but we don't support this in our design system, yet.

joe-s-avaya:
> I don't believe tooltips have a clickable action as they are typically displayed on hover. There have been discussions on a variant where the tooltip would stay open as there are clickable actions in the tooltip copy, but we don't support this in our design system, yet.

@matthewAvaya, correct, this is not a clickable area, it's responding to the Escape key press on the keyboard as that is a a11y requirement at described in [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role), [W3](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/), and [Deque](https://dequeuniversity.com/library/aria/tooltip).

matthewAvaya:
That's interesting. I learned something new today! If that's the case, the ESC key should remove the tooltip on all the examples, including on hover, rather than just the button example in Storybook.

src/components/Tooltip/Tooltip.tsx
enrique-prado: I think this is good enough. We can iterate if we get bug reports.

src/components/Modal/BasicModal/BasicModal.tsx
```diff
@@ -35,8 +36,8 @@ export const BasicModal = forwardRef(
     const buttons = "actions" in rest ? rest.actions : null;
 
     const onKeyDown = useCallback(
-      (e: { key: string }) => {
-        if (e.key === "Escape" && open) {
+      (e: KeyboardEvent) => {
+        if (e.key === Keys.ESC && open) {
```
enrique-prado, line 40: This seems unrelated to the tooltip logic. I assume it's just refactoring to use the KeyboardEvent object.

enrique-prado (approved):

src/components/Modal/BasicModal/BasicModal.tsx
joe-s-avaya: correct

PR #85, opened 2022-12-01
Added prop to override internal setting of active state in TopLinkItem

mrazian85:

While doing the work to re-factor the Left Navigation on the dpv3 from an Astro Component back into a React Component, I encountered an issue whereby the way in which the Component internally sets which link should be active was causing a delayed visual effect when the Component is re-rendered after navigating to a new page in Astro.

After conferring with @enrique-prado, we decided the best solution was to provide a prop to override the internal Left Navigation logic and allow for custom class names to be passed in instead, thus providing a way to bypass setting active state via `useEffect`, and allowing for a smoother visual effect on render. This PR is my implementation of this idea.

Left Nav without Override:

Left Nav without Override

Left Nav with Override:

Left Nav with Override


joe-s-avaya:
you deleted all the badges? that's a no-no

src/components/LeftNav/TopLinkItem/TopLinkItem.tsx
```diff
@@ -50,7 +53,8 @@ export const TopLinkItem = ({
     <li
       className={clsx(
         "neo-leftnav__main",
-        isActive && "neo-leftnav__main--active"
+        isActive && "neo-leftnav__main--active",
+        className
```
enrique-prado, line 57: Looks good but you'll need to add similar logic to the LinkItem component (which are inside categories) But you can do that as a separate ticket.

enrique-prado (approved):
Left one comment.

mrazian85:
> you deleted all the badges? that's a no-no

Oh whoops. I did not catch that. Let me see what went wrong there

mrazian85:
@joe-s-avaya badges are back! And 0.1% higher coverage ✨

src/components/LeftNav/TopLinkItem/TopLinkItem.tsx
mrazian85: Thanks Enrique!

Thought it looked to me like the `LinkItem` can already accept an `active` prop. I assumed we should already be able to override the active state that way? It also accepts a `className` prop

enrique-prado: Ok, that's perfect. You can ignore my comment then.

joe-s-avaya (approved):
this feels real hacky to me, but I don't have an improvement to suggest, so I'll approve, but it feels like something we may want to re-work in the future

PR #89, opened 2022-12-07
NEO-1250 select component aria compliance update

yangAtSpoken:
[Please check Select component](https://deploy-preview-89--neo-react-library-storybook.netlify.app/?path=/story/components-select--basic-selects)

This work is to update Select component to pass axe-core 4.5 rules.  This goal is achieved by removing checkbox + label combo in InternalSelectOption and by using a custom option component, OptionWithCheckbox.

**IMPORTANT NOTES:**
- Also fixed double down chevron problem in production as shown here in MultiSelect.
- Had to remove the one from the toggle button and left the one from span because it is positioned correctly.

Before:
Screen Shot 2022-12-07 at 2 18 44 PM

Fixed:
Screen Shot 2022-12-07 at 2 20 27 PM


src/components/Select/InternalComponents/MultiSelect.tsx
```diff
@@ -57,7 +57,7 @@ export const MultiSelect = () => {
 
         <button
           {...getToggleButtonProps()}
-          className="neo-multiselect__header"
+          className="neo-multiselect__header neo-multiselect__header--no-after"
```
mrazian85, line 60: 👍

src/components/Select/InternalComponents/OptionWithCheckbox_shim.css
```diff
@@ -0,0 +1,85 @@
...
+  font-size: 11px;
+  display: flex;
+}
+
+.neo-option.neo-option--disabled {
+  color: var(--checkbox-disabled-font-color);
+  cursor: not-allowed;
+}
+
+.neo-option.neo-option--disabled:before {
```
mrazian85, line 65: It seems the `disabled` border style is not being applied correctly; you may have to increase the specificity of this selector, for example:

```suggestion
option.neo-option.neo-option--disabled:before {
```

yangAtSpoken: Thx for pointing this out. In the end, had to add .neo-option again to win the specificity battle.

matthewAvaya:
Basic Selects
- Remove the text 'Select One' as it's unnecessary as the label and helper text instructs the user what to do. **Applies to all versions.**
- On the first select input, the dropdown doesn't scroll when using the keyboard to navigate the list.
- On the second select input, I can't open it by clicking on the chevron icon like it does on the one above it. 
- Let's add some spacing between the two select components for better separation. **Applies to all versions.**
- The select input field with the chips has unnecessary empty space. Is there a reason for the empty space below the chips? **Applies to all versions.**
- If dropdown is open, chevron icon should be facing up. **Applies to all versions.**

Searchable
- Remove placeholder text. **Applies to all versions.**

Required in form
- Favorite food select looks like it's in disabled mode and yet I can close 'Blueberries'. I would expect that everything would be disabled. 

Empty
- Do we need this option if there are no selections provided? 

Native select - form submission
- Make the Reset button to use the secondary button (outlined)

All
- The hover states on the dropdown looks odd as it overrides the focus. **Applies to all versions.**

yangAtSpoken:
> Please check Select component

@matthewAvaya   Could you double check with production for the same issues you laid out?  I will fix regressions.  Otherwise, I will create an improvement ticket.  The current ticket is already 5 as it is. Here is the link to production, https://neo-react-library-storybook.netlify.app/?path=/story/components-select--basic-selects

yangAtSpoken:
Created https://jira.forge.avaya.com/browse/NEO-1350 to track improvements proposed by @matthewAvaya .

PR #103, opened 2023-01-03
NEO-1376 // add default to `aria-label`, update TS+error-message to force use of `aria-label*` (XOR)

joe-s-avaya:
[Link to updated components in Deploy Preview](UPDATEMEWITHALINK)

Use "Hide whitespace" when looking at "Files changed"
Screenshot 2023-01-03 at 1 37 05 PM

As stated in the previous PR, we were already enforcing the use of `aria-label`, so this is just adding a default and telling the user that they _must_ pass one (and only one, ie, XOR) of `aria-label` or `aria-labelledby`.


src/components/LeftNav/LeftNav.test.tsx
```diff
@@ -1,103 +1,140 @@
...
-    let renderResult: RenderResult<
-      typeof import("@testing-library/dom/types/queries"),
-      HTMLElement,
-      HTMLElement
-    >;
-
-    beforeEach(() => {
-      renderResult = render(<Default />);
+describe("LeftNav", () => {
+  describe("base tests", () => {
```
yangAtSpoken, line 14: Call this a11y tests, perhaps?  And add a positive test case where aria-labelledby is used.

joe-s-avaya: 🤦  good call on both accounts, will update, thanks!

src/components/LeftNav/LeftNav.test.tsx
```diff
@@ -1,103 +1,140 @@
...
-    beforeEach(() => {
-      renderResult = render(<CategoryGroups />);
-    });
-
-    it("should render ok", () => {
-      const { container } = renderResult;
-      expect(container).not.toBe(null);
+  describe("storybook tests", () => {
+    describe("Default", () => {
+      let renderResult: RenderResult<
```
yangAtSpoken, line 49: nit: extract RenderResult type out like this:

type Queries = typeof import("@testing-library/dom/types/queries");
type RenderResultType = RenderResult<
  Queries,
  HTMLElement,
  HTMLElement
>;

And this line could just be:

    let renderResult: RenderResultType;

yangAtSpoken (approved):
A couple of nits.

src/components/LeftNav/LeftNav.test.tsx
joe-s-avaya: updated!

src/components/LeftNav/LeftNav.test.tsx
joe-s-avaya: Ooh! Yes! I was in the habit copy-pasting and didn't even think of that.

I've centralized the type and removed all copy-pasta (using find-all).

PR #109, opened 2023-01-10
remove dep and copy-paste it's contents into `utils`

joe-s-avaya:

Removed the dep `react-node-to-string` and copy-pasted its contents into the `utils` folder. It came with tests, so I popped those in there as well. 

I ran `yarn all`, but that's really all the testing that I did as our existing tests should take care of any potential issues.


src/utils/react-node-to-string/index.ts
```diff
@@ -0,0 +1,19 @@
+// mostly copy-pasted from: https://github.com/sunknudsen/react-node-to-string
```
joe-s-avaya, line 1: I copy pasted and tweaked it to use our preferred code standards (`export const` instead of `export default`, ect.)

mrazian85 (approved):
Nice

PR #113, opened 2023-01-13
NEO-1388 style changes

yangAtSpoken:
[Select](https://deploy-preview-113--neo-react-library-storybook.netlify.app/?path=/story/components-select--basic-selects)

fixed 3 issues mentioned in the ticket:
* Spacing issues on all multiple select versions. When there are multiple chips selected, the top padding should match the version that only has a single chip.
* On multiple select, the cursor should be positioned next to the last chip rather than on a new separate line.
* hover state should not cut off focus blue border.

I have also forced focus visible on mouse click, as I want to a consistent look between mouse and keyboard actions.  Let me know if this is ok.

### DIFF
--- a/src/components/Select/Select_shim.css
+++ b/src/components/Select/Select_shim.css
@@ -130,7 +130,7 @@ div.neo-multiselect.neo-multiselect--small span.neo-multiselect-combo__header {
   padding: 0;
 }
 
-.neo-multiselect-combo__header input.neo-input__height-26px {
+.neo-multiselect-combo__header input.neo-input--height-26px {
   min-height: 26px;
   height: 26px;
   padding-left: 0;
--- a/src/components/Select/Select_shim.css
+++ b/src/components/Select/Select_shim.css
@@ -12,7 +12,7 @@
 body .neo-multiselect__content.neo-set-keyboard-focus,
 body .neo-multiselect__content:focus-visible {
   outline-offset: 1px;
-  outline: 1px auto #1473e6;
+  outline: 1px auto var(--global-outline-color);
 }
 .neo-multiselect__content:empty {
   border: none;
@@ -78,13 +78,12 @@ div.neo-multiselect button.neo-multiselect__header {
   flex-grow: 1;
 }
 
-span.neo-multiselect-combo__header .neo-padded-container {
+span.neo-multiselect-combo__header .neo-multiselect__padded-container {
   align-items: center;
   background-color: var(--input-background-color);
   color: var(--input-font-color);
-  border: 0px solid var(--input-border-color);
+  border: 0;
   letter-spacing: 0;
-  border-radius: 2px;
   flex-wrap: wrap;
   justify-content: flex-start;
   min-height: 36px;
@@ -96,7 +95,7 @@ span.neo-multiselect-combo__header .neo-padded-container {
 
 .neo-multiselect.neo-multiselect--disabled
   .neo-multiselect-combo__header
-  .neo-padded-container {
+  .neo-multiselect__padded-container {
   background-color: var(--input-disabled-bg);
   color: var(--input-disabled-color) !important;
   cursor: not-allowed !important;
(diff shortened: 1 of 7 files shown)


matthewAvaya (approved):

src/components/Select/Select_shim.css
```diff
@@ -9,7 +9,11 @@
   button.neo-multiselect__header.neo-multiselect__header--no-after:after {
   display: none;
 }
-
+body .neo-multiselect__content.neo-set-keyboard-focus,
+body .neo-multiselect__content:focus-visible {
+  outline-offset: 1px;
+  outline: 1px auto #1473e6;
```
mrazian85, line 15: Suggfestion: Can use Neo global outline color here

```suggestion
  outline: 1px auto var(--global-outline-color);
```

src/components/Select/Select_shim.css
```diff
@@ -65,11 +69,39 @@
...
-  width: 100%;
   min-height: 36px;
 }
+
+div.neo-multiselect button.neo-multiselect__header {
+  min-height: 36px;
+  flex-grow: 1;
+}
+
+span.neo-multiselect-combo__header .neo-padded-container {
```
mrazian85, line 81: Suggest renaming this class to indicate its use specifically with `Select`, something like `.neo-multiselect__padded-container`, unless you intend for it to be used elsewhere

src/components/Select/Select_shim.css
```diff
@@ -65,11 +69,39 @@
...
+  align-items: center;
+  background-color: var(--input-background-color);
+  color: var(--input-font-color);
+  border: 0px solid var(--input-border-color);
```
mrazian85, line 85: It seems both `border` and `border-radius` will have no effect here as the former is set to `0px`

src/components/Select/Select_shim.css
```diff
@@ -65,11 +69,39 @@
...
+  min-height: 36px;
+  padding-top: 3px;
+  padding-bottom: 3px;
+  flex-grow: 1;
+  display: flex;
+}
+
+.neo-multiselect.neo-multiselect--disabled
+  .neo-multiselect-combo__header
+  .neo-padded-container {
```
mrazian85, line 99: Same comment as for line 81 above

src/components/Select/Select_shim.css
```diff
@@ -89,6 +121,22 @@ div.neo-multiselect.neo-multiselect--small span.neo-multiselect-combo__header {
   padding-bottom: 0;
 }
 
+.neo-multiselect
+  span.neo-multiselect-combo__header
+  button.neo-multiselect__header.neo-button--width-10px {
```
mrazian85, line 126: Suggest an alternate implementation in order to avoid overly-verbose class names 

```suggestion
  button.neo-multiselect__header:last-child {
```

src/components/Select/Select_shim.css
```diff
@@ -89,6 +121,22 @@ div.neo-multiselect.neo-multiselect--small span.neo-multiselect-combo__header {
...
+  width: 10px;
+  min-height: 30px;
+  height: 30px; /* same as chip */
+  padding: 0;
+}
+
+.neo-multiselect-combo__header input.neo-input--height-26px {
```
mrazian85, line 133: Similar to line 126 above, can be something like:

```suggestion
.neo-multiselect-combo__header span input:first-child {
```

mrazian85 (commented):
Some comments regarding CSS

src/components/Select/Select_shim.css
yangAtSpoken: sure. thx.

src/components/Select/Select_shim.css
yangAtSpoken: great

mrazian85 (approved):

src/components/Select/Select_shim.css
yangAtSpoken: done.

PR #115, opened 2023-01-16
NEO-1390: Select focus order

mrazian85:
[Link to deployment preview](https://deploy-preview-115--neo-react-library-storybook.netlify.app/?path=/story/components-select--basic-selects)

This PR is for work related to [NEO-1390](https://jira.forge.avaya.com/browse/NEO-1390), whereby the tab focus order on Select elements with Chips is incorrect as per @matthewAvaya. The gist of the solution is to re-order the structure of the elements within the Select while maintaining their visual appearance using the `order` CSS property value


src/components/Select/InternalComponents/MultiSelect.tsx
```diff
@@ -56,16 +56,15 @@ export const MultiSelect = () => {
...
-
           <button
             {...getToggleButtonProps()}
             className="neo-multiselect__header neo-multiselect__header--no-after neo-button--width-10px"
             type="button"
             aria-label={ariaLabel}
-          >
-            {selectedItemsAsChips ? <>&nbsp;</> : placeholder}
-          </button>
+          ></button>
```
mrazian85, line 64: Changing order of elements to effect desired tab order as per [NEO-1390](https://jira.forge.avaya.com/browse/NEO-1390). Visual order of elements is preserved using CSS `order` property in `Select_shim.css` file

src/components/Select/InternalComponents/MultiSelect.tsx
```diff
@@ -56,8 +56,6 @@ export const MultiSelect = () => {
     >
       <span className="neo-multiselect-combo__header">
         <span className="neo-padded-container">
-          {selectedItemsAsChips}
```
mrazian85, line 59: Changing order of elements to effect desired tab order as per [NEO-1390](https://jira.forge.avaya.com/browse/NEO-1390). Visual order of elements is preserved using CSS order property in Select_shim.css file

matthewAvaya:
Tab order looks good. 

Two new issues that I'm seeing:

- Under Basic Selects, I noticed that the focus goes to the blank area next to the chip when browsing with my keyboard while using screen reader (see attached). It also reads the label 'Select a few nice foods' twice. I would expect the focus would be around the entire select component. 
Screen Shot 2023-01-17 at 11 00 09 AM
- When a preselected item, such as an apple, is hovered from the dropdown menu, it says 'apple not selected' even though it's already selected.

src/components/Select/Select_shim.css
```diff
@@ -174,3 +181,7 @@ div.neo-multiselect div.neo-input-group:hover {
   .neo-input-hint:before {
   content: "";
 }
+
+span#multi-searchable-select-toggle-button span > div:first-child {
```
yangAtSpoken, line 185: Where is the ID used?

yangAtSpoken (commented):
one question

src/components/Select/InternalComponents/MultiSelect.tsx
yangAtSpoken: great solution. hope no more requests like this from @matthewAvaya as it is confusing when document order is different from visual order. lol:

src/components/Select/Select_shim.css
yangAtSpoken: looks like this id depends on the id given to Select by dev. So this will not work if a different id is passed to Select.

yangAtSpoken:
> Tab order looks good.
> 
> Two new issues that I'm seeing:
> 
> * Under Basic Selects, I noticed that the focus goes to the blank area next to the chip when browsing with my keyboard while using screen reader (see attached). It also reads the label 'Select a few nice foods' twice. I would expect the focus would be around the entire select component.
> 
> Screen Shot 2023-01-17 at 11 00 09 AM
> 
> - When a preselected item, such as an apple, is hovered from the dropdown menu, it says 'apple not selected' even though it's already selected.

For the first issue, I do not think the whole select could get keyboard focus as it is not an interactive element.  If it were, you would not be allowed to put close buttons inside it.

mrazian85:
> Tab order looks good.
> 
> Two new issues that I'm seeing:
> 
> * Under Basic Selects, I noticed that the focus goes to the blank area next to the chip when browsing with my keyboard while using screen reader (see attached). It also reads the label 'Select a few nice foods' twice. I would expect the focus would be around the entire select component.
> 
> Screen Shot 2023-01-17 at 11 00 09 AM
> 
> - When a preselected item, such as an apple, is hovered from the dropdown menu, it says 'apple not selected' even though it's already selected.

Hey @matthewAvaya, I'm having a look at this now, it looks like there are a few things going on here:

1) With regards to the focus being on the internal element rather than the entire Select component -- as @yangAtSpoken has already commented, I believe that this is necessary for correct accessibility because of the way the Select is designed. Because the Chips appear inside the Select, and they have associated Button elements, this means the parent Select cannot itself be an interactive element (no nested interactive elements allowed).

2) I do see that the 'Select a few nice foods' text is being read twice; this is because the screen reader is reading both the `label`, and then, because we use a `button` as the interactive element for the Select, it's also reading the `aria-label` of the Button. Not sure how to fix this without breaking accessibility...maybe we could give the `aria-label` unique text? In any case, [NEO-1389](https://jira.forge.avaya.com/browse/NEO-1389) has work related to Select elements being announced twice. I suggest this issue be added there.

3) With regards to the 'not selected' issue, I do see that there is a bug/behaviour in the Select where Options are toggled as unselected on mouse out. I don't have a clear idea about how to solve this yet, it will probably require some structural changes to the Select so best to add that work to a separate ticket.

src/components/Select/Select_shim.css
mrazian85: Good catch! I have changed it to use class-based and positional selectors instead

yangAtSpoken (approved):

matthewAvaya:
@yangAtSpoken @mrazian85 Read your notes above and I'm good with where you've landed. I would like to bring these topics to next week's accessibility session with Terri to get her input. Thanks!

mrazian85:
> @yangAtSpoken @mrazian85 Read your notes above and I'm good with where you've landed. I would like to bring these topics to next week's accessibility session with Terri to get her input. Thanks!

Thanks @matthewAvaya; I'll go ahead and add the relevant notes to existing tickets/create new ones; I'll also leave a note in the Slack channel for the session with Terri

PR #116, opened 2023-01-18
dux-1464: change padding-right on accordion header

yangAtSpoken:
[Link to updated components in Deploy Preview](UPDATEMEWITHALINK)

- [ ] reviewed my code changes
- [ ] ensured that all tests pass
- [ ] updated the link at the top of this PR (or remove it if not applicable)
- [ ] added any important information to this PR (description, images, GIFs, ect.)
- [ ] tagged Matt if any visual changes have occurred
- [ ] done an accessibility check on my work

Put the most important information here. The content that any reviewer should see and read first (after the PR title).

**EXAMPLE IMAGE(S):**
[PUT EXAMPLES HERE, such as an image of the new tests, or important use case added by this update]

[PUT EXAMPLE GIF HERE]

**IMPORTANT NOTES:**
- Put any secondary, but still important, notes here
- May be bullets or a single sentence/paragraph

### DIFF
--- a/src/components/Accordion/Accordion_shim.css
+++ b/src/components/Accordion/Accordion_shim.css
@@ -1,0 +1,3 @@
+.neo-accordion__header > button.neo-accordion__header-text {
+  padding-right: 50px !important;
+}
--- a/src/components/Accordion/Accordion.stories.tsx
+++ b/src/components/Accordion/Accordion.stories.tsx
@@ -11,11 +11,15 @@ export const Single = () => (
   <Sheet title="Accordion Examples" style={{ width: 400 }}>
     <Accordion header="Accordion Header">Some text in body</Accordion>
 
-    <Accordion header="Accordion with expand true" defaultExpanded>
+    <Accordion header="Accordion with expand true " defaultExpanded>
       Some text in body
     </Accordion>
 
     <Accordion header="Header Disabled" disabled />
-    <Accordion header="Accordion Header test">Some text in body</Accordion>
+
+    <Accordion header="Lorem ipsum dolor sit amet consectetur, adipisicing elit. Placeat harum in, ut possimus, porro accusamus beatae ab officiis quae consectetur quia? Fugiat ea, sunt fuga numquam accusamus nemo odio praesentium.">
+      Some text in body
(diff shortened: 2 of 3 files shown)


matthewAvaya:
@yangAtSpoken Was there a specific issue with the padding-right on the accordion header? I can't recall what that was.

src/components/Accordion/Accordion_shim.css
```diff
@@ -0,0 +1,3 @@
+.neo-accordion__header > button.neo-accordion__header-text {
+  padding-right: 50px !important;
```
enrique-prado, line 2: Is there another way to fix this without using !important? The selector is fairly specific so I'm not too worried about it so I'll approve it, but if there is an easy alternate fix we should use that instead.

enrique-prado (approved):
Just the one question in case there is a better way to fix this.

src/components/Accordion/Accordion_shim.css
yangAtSpoken: > Is there another way to fix this without using !important? The selector is fairly specific so I'm not too worried about it so I'll approve it, but if there is an easy alternate fix we should use that instead.

The original rule this one is overwriting has an !important in it, so I have to use !important here as well.  Let me know if there is an alternative.

yangAtSpoken:
> @yangAtSpoken Was there a specific issue with the padding-right on the accordion header? I can't recall what that was.

@matthewAvaya You were saying text is is too close to the chevron in https://jira.forge.avaya.com/browse/DUX-1464.

src/components/Accordion/Accordion_shim.css
joe-s-avaya: great point @enrique-prado, and @yangAtSpoken, looks like you are correct, that's the only way to resolve the issue without editing the css-lib

so at least for now, this looks good to me

joe-s-avaya (approved):

src/components/Accordion/Accordion_shim.css
mrazian85: Nvm, found it. It is in the`_list.scss` file. Yes, this is a good way to solve this for now.

PR #125, opened 2023-01-27
NEO-1371 link component

yangAtSpoken:
[Link to Link in Deploy Preview](https://deploy-preview-125--neo-react-library-storybook.netlify.app/?path=/story/components-link--misc)

Implemented Link component and added tests.


src/components/Link/Link.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+        </Link>
+      </li>
+      <li>
+        <Link href="#main" icon="print" placement="right" disabled>
+          Standalone + right Icon + disabled
+        </Link>
+      </li>
+      <li>
+        <Link href="#inline" inline disabled>
+          Inline
```
joe-s-avaya, line 53: Recommend wrapping the two "inline" links in lorem ipsum text so that it's a better example of how they would really look

src/components/Link/Stories_shim.css
```diff
@@ -0,0 +1,3 @@
+ul li {
+  padding: 1rem;
+}
```
joe-s-avaya, line 3: You should use class names, else these styles are going to mess with the styles in other pages. But they'll only mess with those styles if you go to this page first, and then other pages, which will make it extra confusing. Thus, please use class names.
```suggestion
ul li.li-storybook-padding {
  padding: 1rem;
}
```

joe-s-avaya: alternate css (because `has` is new and shiny):
```suggestion
ul li:has(.neo-link) {
  padding: 1rem;
}
```

joe-s-avaya (approved):
minor nit on a story, and a recommendation for a css change to protect from style pollution

src/components/Link/Link.stories.tsx
yangAtSpoken: sure.

matthewAvaya (approved):

mrazian85 (approved):

PR #127, opened 2023-01-27
renovate // update config: pin deps, auto assign devs to review PRs

joe-s-avaya:
Mo's discovered a failing test. After some investigation between him and I, it was only discoverable after we killed our node_modules folder and re-installed deps. I believe that if we "pin" all of our deps, we won't run into an issue like that anymore.


mrazian85 (approved):
Nice; seems like a good solution 👍

PR #140, opened 2023-02-10
neo-1459: update to node 18 and fix storybook scripts

yangAtSpoken:
[Link to updated components in Deploy Preview](https://deploy-preview-140--neo-react-library-storybook.netlify.app/)

Update node to 19.6.0; 

Note: dev needs to have node 19.6.0 locally installed.


Barbacoa08 (changes requested):
We should be using LTS IMO, [which is 18](https://nodejs.org/en/).

.nvmrc
```diff
@@ -0,0 +1 @@
+19.6.0
```
Barbacoa08, line 1: I've not seen a `.nvmrc` file before. I did a quick search and I think that this is simply telling NVM to use Node `v19.6.0`. Is that correct?

yangAtSpoken: Used by netlify

package.json
```diff
@@ -79,6 +79,8 @@
     "@storybook/addon-interactions": "6.5.16",
     "@storybook/addon-links": "6.5.16",
     "@storybook/builder-vite": "0.4.0",
+    "@storybook/builder-webpack5": "^6.5.16",
+    "@storybook/manager-webpack5": "^6.5.16",
```
Barbacoa08, line 83: nit: pin these versions. Mend will do it automatically when it does its weekly check, so this is very nitpicky, but it feels like a good habit to have.

yangAtSpoken:
> We should be using LTS IMO, [which is 18](https://nodejs.org/en/).

Renovate was using the latest node. if we use 18 here, it will update to 19.6.0 next chance it gets.

https://github.com/avaya-dux/neo-react-library/pull/137

.nvmrc
joe-s-avaya: it does _not_ affect local machines then?

joe-s-avaya:
> > We should be using LTS IMO, [which is 18](https://nodejs.org/en/).
> 
> Renovate was using the latest node. if we use 18 here, it will update to 19.6.0 next chance it gets.
> 
> #137

To keep that from happening, we can use the `engines` field in `package.json` ([Mend docs](https://docs.renovatebot.com/node/))

.nvmrc
yangAtSpoken: works on my local with or without it.

joe-s-avaya: My question is not if it works, but what it's purpose is. Is it's purpose to _only_ affect what version of Node that Netlify uses? Or is its purpose to affect every machine, including my machine when I run commands locally?

And then, is it doing that successfully?

yangAtSpoken: It tells Netlify to use the version specified in the file.  Locally, it affects only nvm commands, such as `nvm use`.

yangAtSpoken: Yes. it works successfully in both cases.

package.json
```diff
@@ -40,7 +40,7 @@
     "format": "yarn prettier --write ./src --loglevel warn",
     "lint": "eslint src",
     "build": "tsc && vite build",
-    "build:storybook": "build-storybook",
+    "build:storybook": "NODE_OPTIONS=--openssl-legacy-provider build-storybook",
```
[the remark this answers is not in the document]
yangAtSpoken: This tells Node (versions higher than 16) to load legacy code which is still needed by Storybook. Otherwise, we see the [digital envelope error](https://www.newline.co/@kchan/how-to-fix-the-error-errorerror0308010cdigital-envelope-routinesunsupported--0f8d3f17)

package.json
yangAtSpoken: done.

joe-s-avaya (approved):

PR #150, opened 2023-03-03
Add icon size prop

mrazian85:
[Link to updated components in Deploy Preview](https://deploy-preview-150--neo-react-library-storybook.netlify.app/?path=/story/components-icon--standard-icon-sm)

Adding logic to toggle icon size class names as per the latest additions to the [Neo CSS library](https://design.avaya.com/whats-new/changelog).

Updated to latest Neo CSS version as part of the above work. Also fixed a failing test that relied on Icon Component snapshots.


package.json
```diff
@@ -68,7 +68,7 @@
     "react-table": "7.8.0"
   },
   "devDependencies": {
-    "@avaya/neo": "3.67.0",
+    "@avaya/neo": "^3.68.0",
```
joe-s-avaya, line 71: revert the adding of the carrot (`^`)

src/components/Icon/Icon.tsx
```diff
@@ -75,7 +76,7 @@ export const Icon: React.FC<IconProps> = ({
       result.push(iconClass);
     }
 
-    return result.join(" ");
+    return result.join(" ").trimStart();
```
joe-s-avaya, line 79: just use `trim` no reason to only get one end
```suggestion
    return result.join(" ").trim();
```

joe-s-avaya: for bonus points, convert this using a `switch` to using `clsx` (which is what we should have done from the start 😬, but hindsight is 20-20)

package.json
mrazian85: ✅

src/components/Icon/Icon.tsx
mrazian85: ✅

src/components/Icon/Icon.tsx
```diff
@@ -1,9 +1,10 @@
+import clsx from "clsx";
 import { useMemo } from "react";
```
joe-s-avaya, line 2: Screenshot 2023-03-06 at 10 33 01 AM

joe-s-avaya (approved):
there's a linting issue, but otherwise all good

PR #168, opened 2023-03-21
add gitstream configs

joe-s-avaya:
copy-pasted from the ones we're using in dpv3, plus the "missing-tests" one


.cm/define-sensitive-code-areas.cm
```diff
@@ -0,0 +1,29 @@
...
+    run:
+      - action: add-reviewers@v1
+        args:
+          team_reviewers: ['dux-devs']
+      - action: set-required-approvals@v1
+        args:
+          approvals: 1
+
+sensitive:
+  - src/layouts/
```
yangAtSpoken, line 28: maybe just src

.cm/define-sensitive-code-areas.cm
```diff
@@ -0,0 +1,29 @@
...
+  - src/components/
```
yangAtSpoken, line 29: package.json ?

yangAtSpoken (commented):
couple of comments

.cm/define-sensitive-code-areas.cm
joe-s-avaya: Just to make sure that we're on the same page, this is the "sensitive" section. If any files in these sections are included in the PR, then it ups the needed approvals from one approver to two approvers. Right now it's set to just the `layout` and `components` files (as those are our reused files). 

Thus, I very much do not want to say _everything_ under `src` is sensitive.

joe-s-avaya: oh wait, I was thinking that this was the dpv3... good point... yeah... all of `src` seems excessive... but I'll do that for now at least and we can update later if need be. updated.

yangAtSpoken (approved):

PR #188, opened 2023-04-13
NEO-1435 Single Select and Multiple Select Should Announce The Pre-Selected Options When Tabbed into

yangAtSpoken:
[Link to updated components in Deploy Preview](https://deploy-preview-188--neo-react-library-storybook.netlify.app/?path=/story/components-select--basic-selects)

neo-1435


src/components/Select/InternalComponents/MultiSelect.tsx
```diff
@@ -46,8 +47,27 @@ export const MultiSelect = () => {
...
+          .join(" and ")
+          .concat(
+            `, ${selectedItems.length} of ${filteredOptions.length} selected`
+          ),
+      };
+    }
+    if (ariaLabel) {
+      return { "aria-label": ariaLabel };
+    }
+    return { "aria-labelledby": ariaLabelledby };
```
joe-s-avaya, line 68: Nit and a concern. 

Concern: this is English dependent. Probably need to add a "translations" prop. You can see how I did it for the Table component in this file:
`src/components/Table/types/TranslationTypes.ts`

joe-s-avaya: **Concern**: this is English dependent. Probably need to add a "translations" prop. You can see how I did it for the Table component in this file:
`src/components/Table/types/TranslationTypes.ts`

joe-s-avaya: **nit**: prefer `if-else` as multiple return statements have a higher cognitive load for groking them (just slightly, thus why this is a "nit"). 

```suggestion
   const result = { "aria-labelledby": ariaLabelledby };
   
    if (selectedItems && selectedItems.length > 0) {
      result = {
        "aria-label": selectedItems
          .map((item) => item.value)
          .join(" and ")
          .concat(
            `, ${selectedItems.length} of ${filteredOptions.length} selected`
          ),
      };
    } else if (ariaLabel) {
      result = { "aria-label": ariaLabel };
    }

    return result;
```

src/components/Select/InternalComponents/SingleSelect.tsx
```diff
@@ -21,8 +24,24 @@ export const SingleSelect = () => {
   const {
     role,
     "aria-activedescendant": ariaActiveDescendant,
+    "aria-labelledby": ariaLabelledby,
     ...restToggleProps
   } = getToggleButtonProps();
+
+  const computedAriaProperty = useMemo(() => {
```
joe-s-avaya, line 31: [same as here](https://github.com/avaya-dux/neo-react-library/pull/188/files#r1165859987), prefer `if-else` over multiple returns

src/components/Select/Select.test.jsx
joe-s-avaya, line 1: 👍

joe-s-avaya (approved):
Two nits, one concern. Nothing that _must_ be done in this PR, but I think that the "concern" should be addressed before the ticket is closed.

src/components/Select/InternalComponents/MultiSelect.tsx
yangAtSpoken: If we need translation, we need to translate these words: 
"and" in line 59
"of" and "selected" in 61.  

But I am not sure we need it. a11y seems english only.

src/components/Select/InternalComponents/SingleSelect.tsx
yangAtSpoken: no, I prefer early return over if-else.

src/components/Select/InternalComponents/MultiSelect.tsx
joe-s-avaya: This isn't an a11y issue, it's a usability issue. Workspaces for example _has_ to translate everything to several different languages.

That being said, now that I've thought about it, you could simply say that the dev has to provide the appropriate translation, and adjust your logic to assume that, if `aria-label` is passed, we should _always_ use that. Then it would be the responsibility of the implementor to take care of all translation issues.

src/components/Select/InternalComponents/SingleSelect.tsx
joe-s-avaya: I'd argue that you aren't _really_ early returning. In a loop or search, yeah, I agree, but this is just two possible situations and a "default". And I'd argue that my proposed `if-else` code is more clear on that point. 

Again, this is just a few lines of code, so it doesn't actually matter. So I'll stop arguing, just wanted to make sure that my logic was clear.

src/components/Select/InternalComponents/MultiSelect.tsx
yangAtSpoken: Are sure Workspaces translate everything for a11y? Maybe they just translate the visual elements.

joe-s-avaya: I would assume/hope that they are. And either way, we need to support it.

mrazian85: I think this would be a good question for our next session with Terri

mrazian85 (approved):
Nothing to add to existing discussion

src/components/Select/InternalComponents/MultiSelect.tsx
joe-s-avaya: I don't understand your reticence to make the suggested change (adjust logic to use `aria-label` when passed). It's a simple change, has no downsides (as far as I can tell), and allows our users to make their products more useful in all circumstances.

yangAtSpoken: Because "aria-label" is passed to the Select component, it is for the component.  Here the "aria-label" is for the toggle button.  So it is not ideal to use it here, thus it should take a backseat.  "aria-labelledby" is the id of the label for the Select component.  So that is not ideal either. But it is what Downshift gives us.

yangAtSpoken:
> When selecting an item under the multiple select component, it reads 'object object selected'. It doesn't announce the actual selected item such as broccoli.

@matthewAvaya Fixed, please check again.

matthewAvaya:
The selection is now being announced. But now I can't deselect the selected item unless I use my mouse to move to another option and move it back to the selected item I want to deselect. Also, when I deselect an item, it says it has been selected. Shouldn't it say something like 'item (banana) has been deselected'?

src/components/Select/InternalComponents/MultiSelect.tsx
joe-s-avaya: Ah, that makes perfect sense, thanks for clarifying for me. Now that I have the proper context, I agree with you. And bringing this up with Terri is a great idea. We _will_ need to allow our users to translate things, but describing the situation to Terri _before_ we try to architect a solution is probably the best way to go about it. 

Great thoughts, thanks for taking the time to walk me through this.

yangAtSpoken:
> The selection is now being announced. But now I can't deselect the selected item unless I use my mouse to move to another option and move it back to the selected item I want to deselect. Also, when I deselect an item, it says it has been selected. Shouldn't it say something like 'item (banana) has been deselected'?

@matthewAvaya Deselect is doable now in multiple select. Please verify.  Per our discussion, there will be no announcement when deselecting an item.

matthewAvaya (approved):

PR #207, opened 2023-05-24
Publish 0.8.8, fix TS "types" export issue

mrazian85:

This PR fixes an issue brought to our attention by the ACS team; they encountered the issue when attempting to use the React library with Vite and Typescript > v5.0.

The gist of the issue is that the new "bundler" value for `moduleResolution` in the `tsconfig` file makes it so that TS expects the path to type definitions to be present in the "types" key inside of `exports { "." { } }` in the library's `package.json`. Otherwise TS throws the following error:

```
Could not find a declaration file for module '@avaya/neo-react'. 'c:/Work_directory/Exploration_and_Learnings/react-typescript-vite/react-ts-vite/node_modules/@avaya/neo-react/dist/avaya-neo-react.mjs' implicitly has an 'any' type.
  There are types at 'c:/Work_directory/Exploration_and_Learnings/react-typescript-vite/react-ts-vite/node_modules/@avaya/neo-react/dist/index.d.ts', but this result could not be resolved when respecting package.json "exports". The '@avaya/neo-react' library may need to update its package.json or typings.ts(7016)
  ```
I've attached a screenshot of the e-mail below.

We've repro'ed the issue and can confirm that this fix addresses it successfully.

Screen Shot 2023-05-24 at 9 59 15 PM

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -89,45 +89,45 @@
...
     "react-dom": "18.2.0",
     "react-roving-tabindex": "3.2.0",
     "react-stopwatch": "2.0.4",
     "react-table": "7.8.0",
     "typescript": "5.0.4",
-    "vite": "4.3.8",
+    "vite": "4.3.1",
     "vite-plugin-dts": "2.3.0",
-    "vitest": "0.31.1"
+    "vitest": "0.30.1"
```
joe-s-avaya, line 130: I think that you didn't fully pull from `main`, else you wouldn't be downgrading packages 🤔

mrazian85: Oh weird; yes it looks like something didn't go right with the merge. Taking a look now 👀

mrazian85: @joe-s-avaya fixed ✅

PR #214, opened 2023-06-13
update all deps

joe-s-avaya:
[Link to updated Notification component](https://deploy-preview-214--neo-react-library-storybook.netlify.app/?path=/story/components-notification--alert-show)

There were a bunch of major updates to packages which caused breaking changes. This updates _everything_ and fixes all issues.

**IMPORTANT NOTES:**
- Lots of config changes to Storybook. Most of it is simplified now, which is awesome.
- The Notification component had some TS issues due to not accepting the `icon` prop when the type is certain things. This appears to be correct, but now there is no way to set an icon for some types of Notifications, which seems wrong. Will create a ticket for this.
- The Image tests are no longer looking for the `img` role but the `presentation` role. This appears to be because vitest isn't actually loading the image before the role check happens, and thus it's registered as a "presentation" element. I did some quick googling and couldn't figure out anymore than that. Will also open a ticket for that issue.
- I removed the `toastLogger` from the Toast stories. Unsure how to add them back in after the config update required from the Storybook update. 
- Several of the storybook command line commands changed, but the script updates in `package.json`.

UPDATE: added stories: [1731](https://jira.forge.avaya.com/browse/NEO-1731), [1732](https://jira.forge.avaya.com/browse/NEO-1732), [1733](https://jira.forge.avaya.com/browse/NEO-1733)


yangAtSpoken (approved):
Yes, Storybook 7.0. Pretty big change.

PR #241, opened 2023-08-15
NEO-1804 PR #1: Notification Component re-work new features

mrazian85:
[Link to updated components in Deploy Preview](https://deploy-preview-241--neo-react-library-storybook.netlify.app/?path=/docs/components-notification--docs)

As per [NEO-1804](https://jira.forge.avaya.com/browse/NEO-1804), this is the first of what will be two PRs in the React library to re-work our Notification Component.

This PR is focused on implementing the new features required by the Notification re-work, as provided for in the [Figma spec](https://www.figma.com/file/i16TU5abQ1IVCk02ARAEJY/Component-Exploration-FY23---v1?type=design&node-id=4754-454226&mode=design&t=f99zOynS0SPaw162-0) and further guidance by @matthewAvaya. These include:

- Re-working the CSS styling of the Notifications
- Allowing for a combination of different actions to be passed in
- Allowing for inline vs non-inline configurations
- Adding new timestamp, badge, and text-truncation features to the non-inline Notification version.

I have commented the code as needed for each of the above.

The next PR will focus on allowing the non-inline Notification to be rendered at the parent-level of the document, as well as re-writing all the tests for this Component (incl. a11y tests). I have commented out all the Notification tests/added a temporary test to prevent the testing suite from throwing errors in the meantime.

**EXAMPLE IMAGE(S):**
Screen Shot 2023-08-14 at 8 32 51 PM

Screen Shot 2023-08-14 at 8 32 46 PM

notification_truncation

**IMPORTANT NOTES:**
- This is a large PR made even larger by the inclusion of all the new CSS styles and Storybook updates. Net new is approx. 300 LOC


src/components/Notification/Notification.tsx
```diff
@@ -64,15 +71,41 @@ export { logger as notificationLogger };
...
 export const Notification = ({
   type,
-  action,
+  actions,
   header,
   description,
   isElevated = false,
+  isInline = true,
+  occurences = 0,
+  locale = "en-US",
```
mrazian85, line 80: I decided that it would be easier to allow the user to pass in the `locale` prop as necessary, rather than use the `navigator` object provided by browsers, as this will also allow for the Component to be rendered server-side where client-side objects might not be available.

src/components/Notification/Notification.tsx
```diff
@@ -64,15 +71,41 @@ export { logger as notificationLogger };
...
+  }).format(timestampFormat);
+
+  const notificationRef = useRef<HTMLDivElement>(null);
+  const titleRef = useRef<HTMLDivElement>(null);
+  const descriptionRef = useRef<HTMLDivElement>(null);
+
+  const [width, setWidth] = useState("");
+
+  useEffect(() => {
+    if (notificationRef.current)
```
mrazian85, line 103: This block of code dynamically sets the width of the sections to be truncated according to the width of the parent Notification element

src/components/Notification/Notification.tsx
```diff
@@ -112,29 +201,36 @@ const createClickHandler = (
...
-    internalAction = <ClosableAction onClick={handler} />;
-  } else if ("count" in action) {
-    internalAction = <CounterAction count={action.count} />;
-  } else if ("buttons" in action) {
-    internalAction = <ButtonAction buttons={action.buttons} type={type} />;
-  } else {
-    const { onClick, ...rest } = action as ClosableActionProps;
-    const handler = createClickHandler(setClosed, onClick);
-    internalAction = <ClosableAction onClick={handler} {...rest} />;
+  let closableAction = null;
```
mrazian85, line 209: Instead of returning just one action, this function now allows several actions to be created instead

src/components/Notification/Notification.tsx
```diff
@@ -64,15 +71,41 @@ export { logger as notificationLogger };
  */
 export const Notification = ({
   type,
-  action,
+  actions,
   header,
   description,
   isElevated = false,
+  isInline = true,
+  occurences = 0,
```
mrazian85, line 79: The user will be responsible for handling the number of occurrences for a single Notification and passing that in accordingly

src/components/Notification/Notification.tsx
joe-s-avaya: unless you _can't_ use `navigator`, I'd set the default to be:
```suggestion
  locale = navigator?.languages[0] ?? "en-US",
```

src/components/Notification/Notification.tsx
```diff
@@ -86,15 +119,71 @@ export const Notification = ({
       <div
         role="img"
         className={clsx("neo-notification__icon", icon && `neo-icon-${icon}`)}
-        aria-label={`icon ${icon}`}
+        aria-label={clsx("icon", icon && icon)}
       />
-      <div className="neo-notification__message">
-        {header && <div className="neo-notification__title">{header}</div>}
+      <div className="neo-notification__message" ref={notificationRef}>
+        <div style={{ display: "flex", alignItems: "center" }}>
```
joe-s-avaya, line 125: don't use `style`, you can fix this in a later PR, but we shouldn't be using the `style` prop in prod code

src/components/Notification/Notification.tsx
```diff
@@ -86,15 +119,71 @@ export const Notification = ({
...
+          {!isInline && (
+            <p className="neo-notification__timestamp">{timestamp}</p>
+          )}
+          {!isInline && occurences > 1 && (
```
joe-s-avaya, line 129: The `occurences` is not expected to be updated inline correct? If it _is_ expected to update in line, then you'll need to put a placeholder here so that sibling items don't shift around if/when this element appears/disappears.

src/components/Notification/Notification.tsx
```diff
@@ -86,15 +119,71 @@ export const Notification = ({
...
+          )}
+          {!isInline && (
+            <IconButton
+              className={clsx(
+                "neo-notification__button",
+                !isTruncated && "neo-notification__button--pressed",
+              )}
+              variant="tertiary"
+              icon="chevron-down"
+              aria-label="expand/collapse"
```
joe-s-avaya, line 143: nit: all `aria-label` must be allowed to be overwritten so that it can be translated when necessary (same on several other lines such as 132)

src/components/Notification/Notification.tsx
```diff
@@ -86,15 +119,71 @@ export const Notification = ({
...
+
+        {header && (
+          <div
+            className={clsx(
+              "neo-notification__title",
+              isTruncated &&
+                "neo-notification__text--truncated neo-notification__title--truncated",
+            )}
+            ref={titleRef}
+            style={isTruncated ? { width } : {}}
```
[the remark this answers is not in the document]
joe-s-avaya: Happy to jump on a call and pair if you want someone to discuss with and maybe write up some codesandbox examples to argue over

src/components/Notification/notification_shim.css
```diff
@@ -0,0 +1,197 @@
...
+}
+
+.neo-notification__timestamp {
+  color: var(--neo-color-base-500);
+  font-size: 0.75rem;
+  margin-right: 0.5rem;
+}
+
+.neo-notification__message {
+  container-type: inline-size;
```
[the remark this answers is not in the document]
joe-s-avaya: hmmmm... great question! I have no idea 😅 

There's a lot of code here, so it's pretty hard to grock. Would you like to pair on a simplier version of this in a codesandbox so that we can test out ideas?

joe-s-avaya (approved):
I have a lot of suggestions, but as a "part 1" this is good. I recommend merging as is, but keeping track of all comments so that future PRs can work on them. 

Or at least that's what I think is the easier route, feel free to resolve in this PR or future PRs. Whichever is easier for you to keep track of.

src/components/Notification/notification_shim.css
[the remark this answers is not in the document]
mrazian85: Yes sure; that's a good idea. Since it seems to be working for the purposes of this PR, I mightn't take you up on that right away. But I will add a note to set some time aside to explore this further with you as I think it could be a good learning experience for us both 🙏

src/components/Notification/Notification.tsx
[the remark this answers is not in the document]
mrazian85: The purpose here is not to force things into a single line, and I think that there isn't another way to make sure the value of `width` can change dynamically based on the size of the Notification -- at least, not using class names.

Am going to pick up some of your other suggestions first and then will ping you to discuss this particular point in more detail.

src/components/Notification/Notification.tsx
mrazian85: Great point. Do you have any example handy of where we do this in other Components? I think I'd like to copy the implementation

src/components/Notification/Notification.tsx
mrazian85: ✅

matthewAvaya:
- Font size is on the date and time stamp should be **web/small Body - Semibold**
- The font color on the date and time stamp should be using **base/600**
- The header text size should be using **web/body - semibold** which is 14 pt, not 19 pt
- The font colors for header and body copy should be using **base/900**
- The font color of the timer should be using **base/900**
- When I resize the component, by dragging the width of the browser, the header and the body text do not wrap as expected (see screenshot).
Screen Shot 2023-08-16 at 10 48 42 AM

src/components/Notification/Notification.tsx
```diff
@@ -64,15 +58,41 @@ export { logger as notificationLogger };
...
+  locale = navigator?.languages[0] ?? "en-US",
   ...rest
 }: NotificationProps) => {
   const icon = "icon" in rest ? rest.icon : null;
   const [closed, setClosed] = useState(false);
-  const internalAction = createAction(action, type, setClosed);
+  const internalActions = createActions(actions, type, setClosed);
+
+  const timestampFormat = Date.now();
+  const timestamp = new Intl.DateTimeFormat(locale, {
```
enrique-prado, line 75: nit: timestampFormat variable name is misleading, currentTime is probably a better name.

enrique-prado (approved):
Just the one nit

matthewAvaya:
@mrazian85  Forgot to also mention that the inline versions of the notification will need the date and time stamps positioned above the header. Thanks!

mrazian85:
@matthewAvaya picked up most of your design feedback as well as the request re: the timestamp on the inline version of the Notification; the text truncation issue on resize I will be solving in the next PR.

Have another look when you get the chance 🙏

src/components/Notification/Notification.tsx
mrazian85: ✅

matthewAvaya (approved):

PR #242, opened 2023-08-22
NEO-1804 PR #2

mrazian85:
[Link to updated components in Deploy Preview](https://deploy-preview-242--neo-react-library-storybook.netlify.app/?path=/docs/components-notification--docs)

This is PR #2 for [NEO-1804](https://jira.forge.avaya.com/browse/NEO-1804), the Notification Component re-work. This PR includes the following changes:

- Added logic to conditionally render truncation toggle based on width of Notification and re-size text appropriately
- Added logic to allow for rendering the Notification at different locations on the screen based on whether it is inline vs not
- Further CSS styling tweaks
- Added logic to pass in default as well as localized values for all `aria-label` attributes in Component where necessary
- Shored up testing file to account for new Component logic

**EXAMPLE IMAGE(S):**
Screen Shot 2023-08-22 at 3 30 56 PM

Truncation toggle behaviour


src/components/Notification/Notification.tsx
```diff
@@ -186,6 +226,10 @@ export const Notification = ({
       </div>
     </div>
   );
+
+  return isInline
+    ? notification
+    : ReactDOM.createPortal(notification, document.body);
```
yangAtSpoken, line 232: We should still use PopupManager to position the notification when it is not inline.  In storybook, we have a couple of stories under PopupManager -> Notification. There is also a hook, usePopup, that simplifies this a bit.

yangAtSpoken (changes requested):
Left a comment.

src/components/Notification/Actions/ButtonAction/SecondaryButton.tsx
```diff
@@ -1,15 +1,18 @@
 import { Button } from "components/Button";
+
 import { SecondaryButtonProps } from "./ButtonActionTypes";
 
 export const SecondaryButton = ({
   children,
+  notificationType,
```
joe-s-avaya, line 7: should this have a default?

src/components/Notification/Notification.tsx
```diff
@@ -98,14 +107,45 @@ export const Notification = ({
...
+  const [needsTruncation, setNeedsTruncation] = useState(false);
+
+  const notificationTranslations = useMemo(() => {
+    return {
+      ...defaultTranslations,
+      ...translations,
+    };
+  }, [translations]);
+
+  function handleResize() {
```
joe-s-avaya, line 120: nit: for consistency I recommend using arrow methods
```suggestion
  const handleResize = () => {
```

src/components/Notification/Notification.tsx
```diff
@@ -77,11 +80,17 @@ export const Notification = ({
...
   ...rest
 }: NotificationProps) => {
   const icon = "icon" in rest ? rest.icon : null;
   const [closed, setClosed] = useState(false);
-  const internalActions = createActions(actions, type, setClosed);
+  const internalActions = createActions(
+    actions,
+    type,
+    setClosed,
+    translations?.closeAction,
```
joe-s-avaya, line 92: your translations all have defaults yes? you shouldn't need the `?`

src/components/Notification/ToggleNotification.stories.tsx
```diff
@@ -28,6 +28,7 @@ export const AlertShow = () => {
         <Notification
           type="alert"
           header="Alert"
+          isInline={false}
```
joe-s-avaya, line 31: nit: maybe have `false` be the default? would that simplify the usage?

joe-s-avaya (approved):
There are some nits and one bug, but I won't block. They're all quick and easy so I assume that you'll take care of it before merging.

matthewAvaya:
Two late updates from the team.
- Make both the header and the body copy to be 2 lines worth before truncation. Applies to both general and inline.
- Increase the padding between the header and the body copy to be 4 px. Applies to both general and inline.

Additional comments: 
- I'm not seeing a tooltip when hovering over the badge.

src/components/Notification/Notification.tsx
mrazian85: Great point; I have restored use of PopupManager to display the notification by moving the render logic outside of the Component itself and updating the stories in `PopupManager.notification.stories.tsx`.

I have also added a default width for the Notification when used with PopUpManager and allowed the user to pass in their own width for custom styling in order to preserve the way text truncation logic is handled

src/components/Notification/Actions/ButtonAction/SecondaryButton.tsx
mrazian85: Good call; I have added one ✅

src/components/Notification/Notification.tsx
mrazian85: The translations do have defaults, but these are handled by the `defaultTranslations` import

The `translations` prop itself is optional

src/components/Notification/Notification.tsx
mrazian85: Actually ended up moving this method out of the Component as I had to refactor the logic to resolve some issues when using the PopUpManager for non-inline Notifications

As an aside, I think that makes this and other methods in the file easier to unit tests, although I haven't added any tests for them in this PR as it is already quite large. I think I will either create a third PR for this, or come back and add these tests after completing the work in the CSS library, as this ticket has already ballooned quite a bit due to changing AC

src/components/Notification/ToggleNotification.stories.tsx
mrazian85: Yes, good call ✅

mrazian85:
> Two late updates from the team.
> 
> * Make both the header and the body copy to be 2 lines worth before truncation. Applies to both general and inline.
> * Increase the padding between the header and the body copy to be 4 px. Applies to both general and inline.
> 
> Additional comments:
> 
> * I'm not seeing a tooltip when hovering over the badge.

@matthewAvaya all feedback should be picked up now, except for the padding between header and body was already at 4px as far as I could tell

matthewAvaya (approved):
Looks great!

yangAtSpoken (approved):

PR #251, opened 2023-09-05
Fixed type issue w/ Timer, fixed Cypress warnings re: excessive linking of methods

mrazian85:
Fixing two issues that are causing a `renovate` PR to fail:

1) A type issue in the `src/components/AgentCard/Timer.tsx` file -- confirmed Component Story still working
2) Fixed linting errors related to chaining multiple methods off a single call to `cy.` in Cypress test file

### DIFF
--- a/src/components/AgentCard/Timer.tsx
+++ b/src/components/AgentCard/Timer.tsx
@@ -6,7 +6,7 @@ export interface TimerProps {
 
 export const Timer = ({ agentStatus }: TimerProps) => {
   const [count, setCount] = useState(0);
-  const timerIdRef = useRef<undefined | NodeJS.Timer>(undefined);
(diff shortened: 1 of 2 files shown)


src/components/AgentCard/Timer.tsx
```diff
@@ -6,7 +6,7 @@ export interface TimerProps {
 
 export const Timer = ({ agentStatus }: TimerProps) => {
   const [count, setCount] = useState(0);
-  const timerIdRef = useRef<undefined | NodeJS.Timer>(undefined);
+  const timerIdRef = useRef<undefined | NodeJS.Timeout>(undefined);
```
joe-s-avaya, line 9: ooh, nice 👍

joe-s-avaya (approved):

PR #256, opened 2023-09-11
Notification `navigator` object fix

mrazian85:
[Link to updated components in Deploy Preview](https://deploy-preview-256--neo-react-library-storybook.netlify.app/?path=/docs/components-notification--docs)

The way the `navigator` object was being checked for in the `locale` prop of the Notification Component was causing the Component to fail when rendered on the server ex. in our current Design Portal.

After linking the React library to my local instance of the DPv3 and trying a few different potential fixes, I've found that simply changing the way the `navigator` object is checked for adequately guards against such errors from occuring.

In addition, I've added some tests to check for both types of behavior (`navigator` defined vs. undefined).

**EXAMPLE IMAGE(S):**

Screen Shot 2023-09-11 at 3 20 15 PM


src/components/Notification/Notification.tsx
```diff
@@ -80,7 +79,7 @@ export const Notification = ({
   isElevated = false,
   isInline = false,
   occurences = 0,
-  locale = navigator?.languages[0] ?? "en-US",
+  locale = typeof navigator !== "undefined" ? navigator.languages[0] : "en-US",
```
yangAtSpoken, line 82: Is this change doing more?

mrazian85: Well, in practice no, but it happens that using either optional chaining or the null coalescing operator was causing this check to fail on the server.

I have a few ideas as to why -- the most convincing is that `vite` does not transpile the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) as per [this docs page](https://main.vitejs.dev/guide/migration.html) (see the "Modern Browser Baseline change" section). I am more or less convinced it is a transpilation issue as changing the syntax -- not the logic -- is what fixed the issue.

yangAtSpoken: The two tests above are testing this line of code, correct? If yes, I would extract the code into a function with navigator as a param.  Then test the function directly.  It would be simpler to and easier to understand.

mrazian85: Yes, that's correct. Yes I can do that. Please go ahead and request changes to this PR and I will ping when it's ready for another look 🙏

yangAtSpoken: > Well, in practice no, but it happens that using either optional chaining or the nullish coalescing operator was causing this check to fail on the server.
> 
> I have a few ideas as to why -- the most convincing is that `vite` does not transpile the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) as per [this docs page](https://main.vitejs.dev/guide/migration.html) (see the "Modern Browser Baseline change" section). I am more or less convinced it is a transpilation issue as changing the syntax -- not the logic -- is what fixed the issue.


If that is the case, still extract the code into a function.  Test the function directly, instead of indirectly by mount the component.

src/components/Notification/Notification.test.jsx
```diff
@@ -32,6 +33,58 @@ const { Event, EventCloseAlert, EventButtons, EventCounter } =
...
+      hour: "2-digit",
+      minute: "2-digit",
+    };
+
+    it("reverts to default value for `locale` prop when navigator object is undefined", () => {
+      const { DateTimeFormat } = Intl;
+
+      vi.spyOn(Intl, "DateTimeFormat").mockImplementation(
+        (locale, options) => new DateTimeFormat(locale, options),
+      );
```
joe-s-avaya, line 50: ooh, good way to mock SSR, nice 👍

src/components/Notification/Notification.test.jsx
```diff
@@ -32,6 +33,58 @@ const { Event, EventCloseAlert, EventButtons, EventCounter } =
...
+
+      let renderResult = render(<Notification />);
+
+      const { container } = renderResult;
+
+      expect(container).toBeDefined();
```
joe-s-avaya, line 56: I'd argue that you don't need any of this. Just call `render` and be done. Your other tests take care of ensuring that the component loads properly. 
```suggestion
      render(<Notification />);
```

src/components/Notification/Notification.test.jsx
```diff
@@ -32,6 +33,58 @@ const { Event, EventCloseAlert, EventButtons, EventCounter } =
...
```
joe-s-avaya, line 79: same here as above
```suggestion
      render(<Notification />);
```

joe-s-avaya (approved):
Yang makes a good point of extracting the testable code. I'm approving as the tests and logic look good to me, but please do take his advice if it makes the most sense.

src/components/Notification/Notification.test.jsx
joe-s-avaya: If Yang's recommendations end up being the better way to test this, these comments will no longer apply.

src/utils/getLocale.ts
```diff
@@ -0,0 +1,2 @@
+export const getLocale: () => string = () =>
```
yangAtSpoken, line 1: I would pass navigator in as a parameter.

src/utils/getLocale.test.ts
```diff
@@ -0,0 +1,21 @@
+import { vi } from "vitest";
+
+import { getLocale } from "./getLocale";
+
+describe("getLocale", () => {
+  it("returns 'en-US' when navigator object is undefined", () => {
+    const localeString = getLocale();
```
yangAtSpoken, line 7: here pass in undefined

src/utils/getLocale.ts
mrazian85: If I am understanding your suggestion, IMO that would make the Component somewhat harder to use, as the user will then have to pass in the `navigator` object themselves

src/components/Table/Table.stories.tsx
```diff
@@ -3,7 +3,16 @@ import { Meta, Story } from "@storybook/react";
 import { useEffect, useMemo, useState } from "react";
 import { Column, ColumnInstance } from "react-table";
 
-import { Chip, Icon, List, ListItem, Menu, MenuItem, SelectNative, Tooltip } from "components";
```
mrazian85, line 6: Formatting

yangAtSpoken (approved):

PR #267, opened 2023-09-29
move logic from dpv3 to neo-react

joe-s-avaya:
By moving the logic for displaying JSX to a static HTML string here, we fix the issue in the DPv3. It's a little weird, but I can't figure out any other way to resolve the breakage. 

I have already tested this in the DPv3 by creating the branch `WIP-astrov3` in the DPv3, and creating the temporary version `1.0.3-beta` here in neo-react. I've gone through every single page in the DPv3 to test, and everything is working as expected. So I just need to merge this, release v1.0.3, and then update the `WIP-astrov3` branch in the DPv3.


yangAtSpoken (approved):
Looks good. Hope this sticks.

PR #269, opened 2023-10-02
allow user to pass a `key` for all Tab components

joe-s-avaya:
as all members of an array must have a `key` prop passed for the react compiler

this was found as part of the updates to DPv3


enrique-prado (approved):
I assume this is the only place key needs to be added since it's an optional prop.

PR #273, opened 2023-10-05
Added an if statement so the page doesnt reload when the current page is clicked.

Brayden-Wachowiak:
[Pagination Storybook](https://deploy-preview-273--neo-react-library-storybook.netlify.app/?path=/docs/components-pagination--docs)

Page no longer reloads if you click the current page on the pagination

### DIFF
--- a/src/components/Pagination/Nodes/NavListItemButton.tsx
+++ b/src/components/Pagination/Nodes/NavListItemButton.tsx
@@ -34,9 +34,10 @@ export const NavListItemButton = ({
           "neo-btn-square",
           `neo-btn-square-${variant} neo-btn-square-${variant}--info`,
         )}
-        onClick={(e)=>{
-          if(!isCurrentPage)
-           onPageChange(e, pageToNavigateTo)
+        onClick={(e) => {
+          if (!isCurrentPage) {
+            onPageChange(e, pageToNavigateTo);
+          }
         }}
         variant={variant}
       >
--- a/src/components/Pagination/Nodes/NavListItemButton.tsx
+++ b/src/components/Pagination/Nodes/NavListItemButton.tsx
@@ -34,7 +34,10 @@ export const NavListItemButton = ({
           "neo-btn-square",
           `neo-btn-square-${variant} neo-btn-square-${variant}--info`,
         )}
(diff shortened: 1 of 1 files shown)


src/components/Pagination/Nodes/NavListItemButton.tsx
```diff
@@ -34,7 +34,10 @@ export const NavListItemButton = ({
           "neo-btn-square",
           `neo-btn-square-${variant} neo-btn-square-${variant}--info`,
         )}
-        onClick={(e) => onPageChange(e, pageToNavigateTo)}
+        onClick={(e)=>{
+          if(!isCurrentPage)
+           onPageChange(e, pageToNavigateTo)
+        }}
```
joe-s-avaya, line 40: This is wrong. The `button` should be disabled and the event shouldn't be propagating up to here.

Brayden-Wachowiak: Matthew doesnt want the button to be disabled because the visual changes of the disabled button are too confusing to the user

Brayden-Wachowiak: We decided on a few CSS changes that change the active and hover state of the current page button to make it less confusing for the user and this change is here so we arent reloading the page if we click the current page button. otherwise i believe it works all the same. The only thing that was added was the if statement.

joe-s-avaya: I think that you should be using the `disabled` attribute and updating the disabled CSS styling. Let's discuss in parking lot. 

If we do decide to go with this solution, you should always wrap your `if`s in curly braces.
```suggestion
        onClick={(e)=>{
          if(!isCurrentPage) {
           onPageChange(e, pageToNavigateTo);
          }
        }}
```

joe-s-avaya (approved):

matthewAvaya:
This may need to be in a separate bug ticket, but I want to point this out here in case it can be resolved with this ticket. When using the screen reader, the page number only says the number. Is there a way to say something closer to "Active, page 1" and "Page 2" for non-active page links? Otherwise, this looks good.

PR #343, opened 2024-03-04
NEO-1074: Add Table row height menu to Table component toolbar

enrique-prado:
[Link to updated components in Deploy Preview](https://deploy-preview-343--neo-react-library-storybook.netlify.app/?path=/story/components-table--editable-data)

This PR implements the Row height dropdown menu selector in the table component. It is visible by default but it can be hidden by setting the showRowHeightMenu prop to false.

When visible, the user can set the table's row height to either "Large", "Medium" or "Small".  I made these 3 strings localizable as well as the aria-label for the menu description.

image

## Summary by CodeRabbit

- **New Features**
	- Introduced customizable row heights in the Table component, including large, medium, and compact options.
	- Added a row height selection menu in the TableToolbar for easy adjustments by users.
- **Enhancements**
	- Integrated new translations for row height options to improve user interface accessibility.
- **Refactor**
	- Updated Table component properties and types to support row height customization.


src/components/Table/Table.tsx
```diff
@@ -81,6 +85,7 @@ export const Table = <T extends Record<string, any>>({
   showPagination = true,
   pushPaginationDown = false,
   showRowSeparator = false,
+  showRowHeightMenu = true,
```
joe-s-avaya, line 88: Recommend having it default to `false` as I see very few tables in the wild that allow me to set their row height.

src/components/Table/Table.tsx
```diff
@@ -198,8 +211,8 @@ export const Table = <T extends Record<string, any>>({
           {...getTableProps()}
           className={clsx(
             "neo-table",
-            rowHeight === "compact" && "neo-table--compact",
-            rowHeight === "medium" && "neo-table--medium",
+            rowHeightValue === "compact" && "neo-table--compact",
+            rowHeightValue === "medium" && "neo-table--medium",
```
joe-s-avaya, line 215: You should allow the dev to set this directly. I'm not sure how useful this is if we only ever use the default value to start.

src/components/Table/helpers/default-data.ts
```diff
@@ -8,10 +8,14 @@ export const translations: ITableTranslations = {
     delete: "Delete",
     edit: "Edit",
     filterColumns: "Filter Columns",
+    large: "Large",
```
joe-s-avaya, line 11: nit: use the `SizeTypeMinimal` type instead of writing your own (from `src/utils/size.ts`)

src/components/Table/types/TranslationTypes.ts
```diff
@@ -7,10 +7,14 @@ export interface IToolbarTranslations extends ITableFilterTranslations {
...
   delete?: string;
   edit?: string;
+  large?: string;
+  medium?: string;
   noDataAvailable?: string;
   refresh?: string;
   rowsPerPage?: string;
   searchInputPlaceholder?: string;
+  selectRowHeight?: string;
+  small?: string;
```
joe-s-avaya, line 17: nit: use the `SizeTypeMinimal` type instead of writing your own (from `src/utils/size.ts`)

src/components/Table/Table.tsx
enrique-prado: I hear you but Matt wanted it showing by default.

src/components/Table/Table.tsx
enrique-prado: good catch. Fixed.

joe-s-avaya (approved):

matthewAvaya:
- The top table row with column headers need to reflect the different sizes that match the rest of the rows. 
- The refresh and preferences icon seem to be in different colors of the blue. They should be the same blue color. 
Screen Shot 2024-03-05 at 9 43 52 AM

- The sizes of the refresh and preferences icons are in different sizes. These should be the same size. 
- In the Preferences dropdown, the selected row height needs a checkmark to let the user know which item is the current selection. 

**Accessibility**

- Preferences dropdown does not select an item using the keyboard.

PR #367, opened 2024-04-23
NEO-2179: Removed problematic react-input-autosize package

enrique-prado:
[Link to updated components in Deploy Preview](https://deploy-preview-367--neo-react-library-storybook.netlify.app/?path=/story/components-select--searchable)

This third party react component was causing build time warnings. It is also 4 years old and not well written so I have replaced it with a vanilla html input tag. 
It required minimal css / html changes and I tested all the stories associated with it without any issues.

## Summary by CodeRabbit

- **Refactor**
	- Removed the `react-input-autosize` package and replaced the `AutosizeInput` with a standard `input` element in the multi-select searchable component.
- **Style**
	- Updated CSS styling by removing the `order` property from the multi-select component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,6 @@
     "loglevel": "1.9.1",
     "react-click-away-listener": "2.2.3",
     "react-focus-lock": "2.12.1",
-    "react-input-autosize": "3.0.0",
     "styled-components": "6.1.8"
   },
   "peerDependencies": {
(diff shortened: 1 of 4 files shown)


joe-s-avaya (approved):
wow, that was even easier than I thought it would be! nice 👍

enrique-prado:
the only dependency that the removed package had was prop-types which we already use in neo react so we are good.

PR #427, opened 2024-06-18
NEO-2146 allow collapse of chips in multi select

yangAtSpoken:
[Story #1](https://deploy-preview-427--neo-react-library-storybook.netlify.app/?path=/story/components-select--collapsed-multi-select)
[Story #2](https://deploy-preview-427--neo-react-library-storybook.netlify.app/?path=/story/components-select--collapsed-narrow-multi-select)

- [x] tagged `@avaya-dux/dux-design` if any visual changes have occurred

ADD PR SUMMARY (should be brief, one sentence if possible)

`@avaya-dux/dux-design`: ADD DESIGN SPECIFIC BRIEF DESCRIPTION
- followed the design portal Select page on collapse chips
- did not implement click on chip per discussion
- 2 stories to check under Select

`@avaya-dux/dux-devs`: ADD DEV SPECIFIC BRIEF DESCRIPTION

- changed Chip to allow forward ref
- added collapse prop to Select
- only MultiSelect supports collapse in this PR per requested by external team

## Summary by CodeRabbit

- **New Features**
	- Improved MultiSelect component with enhanced chip rendering and visibility calculations.
	- Introduced logging for better monitoring of chip rendering stages.
- **Tests**
	- Added test suites to validate calculations of chip widths and visibility within containers.

### DIFF
--- a/src/components/Chip/Chip.stories.tsx
+++ b/src/components/Chip/Chip.stories.tsx
@@ -1,5 +1,6 @@
 import type { Meta, Story } from "@storybook/react";
 
+import { useEffect, useRef, useState } from "react";
 import { Chip, type ChipProps, ChipsContainer } from "./";
 
 export default {
@@ -20,6 +21,49 @@ export const Closable = () => (
 		</Chip>
 	</ChipsContainer>
 );
+
+export const WithRef = () => {
+	const chipRef = useRef<HTMLDivElement>(null);
+
+	const [chipWidth, setChipWidth] = useState(0);
+
+	useEffect(() => {
+		if (chipRef.current) {
+			setChipWidth(chipRef.current.offsetWidth);
+		}
+	}, []);
+
+	return (
+		<ChipsContainer>
+			<Chip ref={chipRef}>With Ref</Chip>
+			<div>Chip Width: {chipWidth}px</div>
+		</ChipsContainer>
+	);
+};
+
+export const WithRefArray = () => {
+	const chipRefs = useRef<HTMLDivElement[]>([]);
+	const [chipWidth, setChipWidth] = useState(0);
+
+	useEffect(() => {
+		if (chipRefs.current[0]) {
+			setChipWidth(chipRefs.current[0].offsetWidth);
+		}
+	}, []);
+
+	const setChipRef = (el: HTMLDivElement | null, index: number) => {
+		if (el) {
+			chipRefs.current[index] = el;
+		}
+	};
+
+	return (
+		<ChipsContainer>
+			<Chip ref={(el) => setChipRef(el, 0)}>With Ref Array</Chip>
+			<div>Chip Width: {chipWidth}px</div>
+		</ChipsContainer>
+	);
+};
 
 export const ChipsContainerExamples = () => (
 	<div>
--- a/src/components/Chip/Chip.stories.tsx
+++ b/src/components/Chip/Chip.stories.tsx
@@ -22,7 +22,7 @@ export const Closable = () => (
 	</ChipsContainer>
 );
 
-export const WithRef = () => {
+export const GetWidth = () => {
 	const chipRef = useRef<HTMLDivElement>(null);
 
 	const [chipWidth, setChipWidth] = useState(0);
@@ -34,23 +34,25 @@ export const WithRef = () => {
 	}, []);
 
 	return (
-		<ChipsContainer>
-			<Chip ref={chipRef} onClick={() => console.log("clicked")}>
-				With Ref
-			</Chip>
-			<div>Chip Width: {chipWidth}px</div>
-		</ChipsContainer>
+		<section>
+			<p>This example shows how to get width of a Chip using a ref.</p>
+			<ChipsContainer>
+				<Chip ref={chipRef} onClick={() => console.log("clicked")}>
+					Chip to measure
+				</Chip>
+				<div>Width in px: {chipWidth}</div>
+			</ChipsContainer>
+		</section>
 	);
 };
 
-export const WithRefArray = () => {
+export const GetWidths = () => {
 	const chipRefs = useRef<HTMLDivElement[]>([]);
-	const [chipWidth, setChipWidth] = useState(0);
+	const [chipWidths, setChipWidths] = useState<number[]>([]);
 
 	useEffect(() => {
-		if (chipRefs.current[0]) {
-			setChipWidth(chipRefs.current[0].offsetWidth);
-		}
+		const widths = chipRefs.current.map((chip) => chip.offsetWidth);
+		setChipWidths(widths);
 	}, []);
 
 	const setChipRef = (el: HTMLDivElement | null, index: number) => {
@@ -60,10 +62,16 @@ export const WithRefArray = () => {
 	};
 
 	return (
-		<ChipsContainer>
-			<Chip ref={(el) => setChipRef(el, 0)}>With Ref Array</Chip>
-			<div>Chip Width: {chipWidth}px</div>
-		</ChipsContainer>
+		<section>
+			<p>
+				This example shows how to get width of Chips using an array of refs.
+			</p>
+			<ChipsContainer>
+				<Chip ref={(el) => setChipRef(el, 0)}>Chip 1</Chip>
+				<Chip ref={(el) => setChipRef(el, 1)}>Chip 2</Chip>
+				<div>Widths in px: {JSON.stringify(chipWidths)}</div>
+			</ChipsContainer>
+		</section>
 	);
 };
 
--- a/src/components/Chip/Chip.stories.tsx
+++ b/src/components/Chip/Chip.stories.tsx
@@ -35,7 +35,9 @@ export const WithRef = () => {
 
 	return (
 		<ChipsContainer>
-			<Chip ref={chipRef}>With Ref</Chip>
+			<Chip ref={chipRef} onClick={() => console.log("clicked")}>
+				With Ref
+			</Chip>
 			<div>Chip Width: {chipWidth}px</div>
 		</ChipsContainer>
 	);
--- a/src/components/Chip/Chip.test.tsx
+++ b/src/components/Chip/Chip.test.tsx
@@ -4,7 +4,7 @@ import { axe } from "jest-axe";
 import { vi } from "vitest";
 
 import { Chip } from "./";
-import { Closable, Default, Templated } from "./Chip.stories";
+import { Closable, Default, Templated, WithRef, WithRefArray } from "./Chip.stories";
 
 describe("Chip", () => {
 	const user = userEvent.setup();
@@ -113,6 +113,44 @@ describe("Chip", () => {
 
 			beforeEach(() => {
 				renderResult = render(<Templated>test</Templated>);
+			});
+
+			it("should render ok", () => {
+				const { container } = renderResult;
+				expect(container).not.toBe(null);
+			});
+
+			it("passes basic axe compliance", async () => {
+				const { container } = renderResult;
+				const results = await axe(container);
+				expect(results).toHaveNoViolations();
+			});
+		});
+
+		describe("WithRef", () => {
+			let renderResult: RenderResult;
+
+			beforeEach(() => {
+				renderResult = render(<WithRef />);
+			});
+
+			it("should render ok", () => {
+				const { container } = renderResult;
+				expect(container).not.toBe(null);
+			});
+
+			it("passes basic axe compliance", async () => {
+				const { container } = renderResult;
+				const results = await axe(container);
+				expect(results).toHaveNoViolations();
+			});
+		});
+
+		describe("WithRefArray", () => {
+			let renderResult: RenderResult;
+
+			beforeEach(() => {
+				renderResult = render(<WithRefArray />);
 			});
 
 			it("should render ok", () => {
--- a/src/components/Chip/Chip.test.tsx
+++ b/src/components/Chip/Chip.test.tsx
@@ -7,9 +7,9 @@ import { Chip } from "./";
 import {
 	Closable,
 	Default,
+	GetWidth,
+	GetWidths,
 	Templated,
-	WithRef,
-	WithRefArray,
 } from "./Chip.stories";
 
 describe("Chip", () => {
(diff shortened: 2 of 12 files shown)


src/components/Select/Select.tsx
```diff
@@ -69,6 +69,7 @@ export const Select = (props: SelectProps) => {
 		placeholder = "",
 		required,
 		searchable = false,
+		collapse = false,
```
joe-s-avaya, line 72: recommend having `collapse` default to true.

src/components/Chip/Chip.test.tsx
```diff
@@ -126,5 +132,43 @@ describe("Chip", () => {
...
+			});
+
+			it("passes basic axe compliance", async () => {
+				const { container } = renderResult;
+				const results = await axe(container);
+				expect(results).toHaveNoViolations();
+			});
+		});
+
+		describe("WithRefArray", () => {
```
joe-s-avaya, line 155: I do not understand why these "with ref" stories and tests are necessary. Is it essential to use collapse logic? If so, does the dev (if they want to do the same as our `Select` need to know how to do this? If both of the above are true, we'll need to figure out how to simplify the logic or, at minimum, document it. As of now, I would have no idea how to implement this.

src/components/Chip/Chip.stories.tsx
```diff
@@ -21,6 +22,51 @@ export const Closable = () => (
 	</ChipsContainer>
 );
 
+export const WithRef = () => {
```
joe-s-avaya, line 25: These stories don't tell a story. I have no idea what their purpose is. You need to either add information to them or remove them. As an example of a story with additional information in it, see the story I wrote for the Table named, [Second Page](https://neo-react-library-storybook.netlify.app/?path=/story/components-table--second-page).

src/components/Select/InternalComponents/MultiSelect.tsx
```diff
@@ -28,23 +42,101 @@ export const MultiSelect = () => {
...
+		}
+
+		const containerWidth = chipContainerRef.current?.offsetWidth || 0;
+		const widths = chipRefs.current.map((ref) => ref.offsetWidth || 0);
+		const chipContents: string[] = chipRefs.current.map(
+			(ref) => ref.textContent || "",
+		);
+
+		const result = calculateWidthsUntilExceed(containerWidth, widths);
+		logger.debug(
```
joe-s-avaya, line 116: I would prefer that you remove all `logger.*` calls. It makes it difficult to read the code. If Enrique doesn't agree, than ignore this request, but for me it makes understanding the code frustrating.

src/components/Select/utils/multiSelectHelper.ts
```diff
@@ -0,0 +1,43 @@
+import log from "loglevel";
```
joe-s-avaya, line 1: nit: recommend making this file (and it's test file) a sibling to `MultiSelect.tsx` as it is only used there

src/components/Select/utils/multiSelectHelper.ts
```diff
@@ -0,0 +1,43 @@
...
+const logger = log.getLogger("multiselect-helper-logger");
+logger.disableAll();
+
+// 120px is the space reserved for the badge chip, the 2 buttons and some padding
+const MIN_CHIP_WIDTH = 120;
+export function calculateWidthsUntilExceed(
```
joe-s-avaya, line 7: If I'm understanding this correctly, I think a better name for this would be:
```suggestion
export function calculateCollapsedChips(
```

src/components/Select/utils/multiSelectHelper.ts
```diff
@@ -0,0 +1,43 @@
...
+	containerWidth: number,
```
joe-s-avaya, line 8: You don't use this, recommend removing

src/components/Select/utils/multiSelectHelper.ts
```diff
@@ -0,0 +1,43 @@
...
+
+		totalWidth += width;
+		index = i;
+		logger.debug(
+			`calculateWidthsUntilExceed 2.1: not exceeding: {totalWidth: ${totalWidth}, index: ${index}}`,
+		);
+	}
+
+	const result = {
+		totalWidth,
```
joe-s-avaya, line 36: You don't use this, recommend removing

src/components/Select/utils/multiSelectHelper.ts
```diff
@@ -0,0 +1,43 @@
...
+		index,
```
joe-s-avaya, line 37: nit: recommend rename for clarity
```suggestion
		startCollapsedChipsIndex,
```

src/components/Select/utils/multiSelectHelper.ts
```diff
@@ -0,0 +1,43 @@
...
+		hiddenCount: widths.length - index - 1,
```
joe-s-avaya, line 38: nit: because we use the term `collapse` as the Select prop, we should remain consistent in our naming and use "collapse" instead of "hidden". I also recommend being explicit in what we are collapsing and, instead of using the generic "count", explicitly stating that we are collapsing chips. 
```suggestion
		collapsedChips: widths.length - index - 1,
```

src/components/Select/utils/multiSelectHelper.test.jsx
```diff
@@ -0,0 +1,48 @@
+import { calculateWidthsUntilExceed } from "./multiSelectHelper";
```
joe-s-avaya, line 1: great tests, self documentation is the best

src/components/Select/InternalComponents/MultiSelect.tsx
```diff
@@ -28,23 +42,101 @@ export const MultiSelect = () => {
...
 	} = useContext(SelectContext);
 
-	const selectedItemsAsChips = useMemo(
-		() =>
-			selectedItems.length
-				? selectedItems.map((item, index) => (
+	logger.debug("1. start");
+
+	const chipContainerRef = useRef<HTMLSpanElement>(null); // Ref for the container of chips
+	const chipRefs = useRef<HTMLDivElement[]>([]); // Array to hold refs for each chip
```
joe-s-avaya, line 48: remove redundant comments, the code it perfectly understandable

src/components/Select/InternalComponents/MultiSelect.tsx
```diff
@@ -28,23 +42,101 @@ export const MultiSelect = () => {
...
+
+	// make sure chips are rendered before calculating their widths in second useEffect
```
joe-s-avaya, line 50: That's a good comment, but could you take all of this logic and put it in a method? It would be much easier to grasp if it were in a simple `renderChips(...)` method. 

You don't need to include this update in this PR, but it would make the code significantly easier to read and maintain if it were possible.

joe-s-avaya (commented):
This is very good, and you've done an excellent job in writing tests and stories to ensure that we are well-documented and safe going forward. 

I do have some improvement recommendations for readability and maintainability. Please review my and coderabbit's comments. 

A major one is that I think you should remove all `logger.*` lines. Having those there makes the reliability poor (for me at least). It's like having footnotes inline.

Finally, I couldn't grasp what you're doing in the Multi Select component. I understand the basic idea from your tests: You're calculating how to display the chips. Easy. However, I cannot understand how exactly you're doing it, so I left a few comments on how to make it more readable and thus maintainable.

src/components/Select/Select.tsx
yangAtSpoken: Don't think so as it does not apply to single selects. And it is not implemented on other multi-selects yet.

src/components/Chip/Chip.test.tsx
yangAtSpoken: These stories are used to prove that we could get the widths of chips by using a ref.  Given the confusion, I have renames these stories to GetWidth and GetWidths. Yes the width property is essential in collapse logic in MultiSelect.

src/components/Chip/Chip.stories.tsx
yangAtSpoken: updated

src/components/Select/InternalComponents/MultiSelect.tsx
yangAtSpoken: They are useful in debugging.

src/components/Select/utils/multiSelectHelper.ts
yangAtSpoken: - No, this util function might be used by other Selects.

src/components/Select/InternalComponents/MultiSelect.tsx
joe-s-avaya: I do not find that to be a convincing argument to let these live forever in the code. Yes, they are useful in debugging, but these are just fancy `console.log` calls. They're too specific to a use-case to be long-lived IMO.

src/components/Select/Select.tsx
joe-s-avaya: ah, good point, resolving

src/components/Select/utils/multiSelectHelper.ts
joe-s-avaya: fair, resolving

src/components/Select/utils/multiSelectHelper.ts
yangAtSpoken: ok

src/components/Select/InternalComponents/MultiSelect.tsx
```diff
@@ -28,23 +42,101 @@ export const MultiSelect = () => {
...
+			`4.3 result ${JSON.stringify({ result, widths, containerWidth, chipContents })}`,
+		);
+
+		if (result.hiddenCount === 0 || selectedItemsAsChips === null) {
+			logger.debug(
+				`4.4 no chips to hide, hiddenCount=${result.hiddenCount}, selectedItemAsChips=${reactNodeToString(selectedItemsAsChips)}`,
+			);
+			return;
+		}
+		const shortenChips = selectedItemsAsChips.slice(0, result.index + 1);
```
enrique-prado, line 126: A comment that explains this block of code would be good. Something like "This is where we display a badge with the count of collapsed chips".  BTW, are 'collapsed', 'shortened' and 'hidden' the same thing? the code seems to use these 3 terms interchangeably and it would be nice to use only one to avoid confusion. I vote for 'collapsed'.

src/components/Select/utils/multiSelectHelper.ts
yangAtSpoken: lastVisibleIndex, i.e. the index of the last chip that is not collapsed.

src/components/Select/utils/multiSelectHelper.ts
yangAtSpoken: I would go with collapsedCount

src/components/Select/utils/multiSelectHelper.ts
enrique-prado: I think this utility function needs a comment at the top that explains what it does because it is not obvious just from the name. It should document its overall purpose, the parameters it takes and its return value.

yangAtSpoken (commented):
Great suggestions

enrique-prado (changes requested):
Left a few comments that I think would make the code more maintainable by other devs who encounter this code for the first time.

src/components/Select/InternalComponents/MultiSelect.tsx
yangAtSpoken: Great comment. changed the code to clarify.

src/components/Select/InternalComponents/MultiSelect.tsx
```diff
@@ -28,23 +42,101 @@ export const MultiSelect = () => {
...
+		}
+		logger.debug({
+			renderCount,
+			chipsCount: selectedItemsAsChips?.length || -1,
+		});
+		if (
+			selectedItemsAsChips === null ||
+			renderCount < selectedItemsAsChips?.length
+		) {
+			logger.debug("4.2 wait for rerendering of chips.");
```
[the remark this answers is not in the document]
yangAtSpoken: There are two rendered from two useEffects.  Here are waiting for the rendering of all chips by the first useEffect.

src/components/Select/utils/multiSelectHelper.ts
joe-s-avaya: code updated, resolving comment thread

src/components/Select/InternalComponents/MultiSelect.tsx
joe-s-avaya: If you have a service that needs monitoring, yeah, but this isn't that. When you have a live service, then for sure, these types of logs are helpful, but since this is not a live service, I argue that this is not helpful past the initial implementation. 
I've also never used the `logger.*` calls myself when debugging an issue, so it's very difficult for me to conclude that they are useful based on the added complexity of more lines of code and my never finding them useful in debugging.

Justin-Avaya:
The "v" icon should be displayed as "^" when the menu is displayed.

Example:
Screenshot 2024-06-18 at 1 47 38 PM

** It should be noted that there is one image (under Anatomy) on the Portal that incorrectly displays the "v" icon, which is incorrect and needs an update. It should also be noted, the source of the error seems to stem from the Figma file which also displays the incorrect chevron direction and will also need to be updated.

<hr/>

I can not seem to type into the select component to search for an item as I thought I should be able to do according to the portal. By allowing the user to click the Select box to access or dismiss the menu, we've removed the user's ability to manually filter as shown on the Portal. Is this intentional? If so, we will need updates to both the Figma file and to the Portal.

Example: 

Screenshot 2024-06-18 at 1 56 22 PM

Justin-Avaya:
**Update**

Combo box (as it is known on the portal) has been located under the story name "Searchable" which demonstrates the ability to search/filter items as expected. We may want to think about renaming this so it aligns better. 
In any case, shouldn't this functionality (searchable/combo-box) also exist when we allow the collapse of chips in multi select?

matthewAvaya:
In the first story, is there room to include a chip for Pear before collapsing the rest in a single number chip? Or is the rule to only have two chips before collapsing the rest in a single number chip?

yangAtSpoken:
@Justin-Avaya fixing "v" needs to be done in a new ticket. This work is limited to regular multi select only; searchable and creatable multi select needs to be done in another ticket.

yangAtSpoken:
> In the first story, is there room to include a chip for Pear before collapsing the rest in a single number chip? Or is the rule to only have two chips before collapsing the rest in a single number chip?

@matthewAvaya impossible at this point without make huge changes to css style.

PR #436, opened 2024-06-26
NEO-2306 // Table bug fix

joe-s-avaya:
[Link to updated Table story "Editable Data"](https://deploy-preview-436--neo-react-library-storybook.netlify.app/?path=/story/components-table--editable-data)

- [ ] ~tagged `@avaya-dux/dux-design` if any visual changes have occurred~

Resolving an issue in the Table where, when the data is removed, `handlePageChange` can cause the component to break. This covers that issue, adds tests for it, and bumps the neo-react version so that I can push this fix to the respective team ASAP.

`@avaya-dux/dux-devs`: test any edge cases for pagination that you can think of. Note that the storybook story tells you when the handler is called and with what data. The actual fix is returning early in the `useEffect` and ensuring that we never return negative page indexes.

## Summary by CodeRabbit

- **New Features**
  - Added page change handling in the `EditableData` component, allowing updates to log items based on page index and size changes.

- **Bug Fixes**
  - Improved the `Table` component to better handle edge cases during page changes, ensuring correct page navigation even when all rows on the last page are deleted.

- **Tests**
  - Added new test cases to verify correct behavior when data is removed, re-added, and when the new data set has fewer pages.

### DIFF
--- a/src/components/Table/Table.test.jsx
+++ b/src/components/Table/Table.test.jsx
@@ -1,5 +1,5 @@
 import { composeStories } from "@storybook/testing-react";
-import { render, screen } from "@testing-library/react";
+import { render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { axe } from "jest-axe";
 import { vi } from "vitest";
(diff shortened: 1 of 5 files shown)


src/components/Table/Table.tsx
```diff
@@ -127,20 +127,19 @@ export const Table = <T extends Record<string, any>>({
...
-			const finalPageIndex = pageCount - 1; // index is 0-based
+			const finalPageIndex = Math.max(0, pageCount - 1); // index is 0-based
 
 			gotoPage(finalPageIndex);
 			setRootLevelPageIndex(finalPageIndex);
 			handlePageChange(finalPageIndex, pageSize);
-		} else if (pageCount > 0 && currentPage === 0) {
-			gotoPage(0);
-			setRootLevelPageIndex(0);
-			handlePageChange(0, pageSize);
```
joe-s-avaya, line 143: after the fix, this became unnecessary

src/components/Table/Table.test.jsx
```diff
@@ -127,6 +127,58 @@ describe("Table", () => {
...
+
+		it("does not update `pageIndex` when all data is removed (refreshed)", async () => {
+			render(<EditableData />);
+
+			const refreshButton = screen.getByLabelText("Refresh");
+
+			expect(screen.getByText("1")).toHaveClass(selectedPageClass);
+
+			// refresh data, removing all data and then adding default data back in
+			await user.click(refreshButton);
```
yangAtSpoken, line 139: Better wrap user action with act() like below:
```suggestion
      import { act } from "@testing-library/react";

      await act(async () => {
        await user.click(refreshButton);
      });
```

yangAtSpoken (commented):
Just need to wrap all user actions in test with an await act().

src/components/Table/Table.test.jsx
joe-s-avaya: Why? I don't think this is necessary and am unsure how to confirm that this is needed or unneeded.

joe-s-avaya: If I run `yarn test Table` I don't see any warnings with the code as is. 
Table test

If I test Select (`yarn test Select`) I do see a warning for using `act`, so they will show up if needed as best I can tell. 
Select test

PR #491, opened 2024-08-13
NEO-2123:  Refactor Table filter as Drawer

enrique-prado:
[Link to Table example with Filter](https://deploy-preview-491--neo-react-library-storybook.netlify.app/?path=/story/components-table--server-side-pagination)
[Link to Figma reference](https://www.figma.com/file/o62kJbRQpFq9T0fGwKvCMA/Component-Exploration-FY22---v3?type=design&node-id=3313-531108&mode=design&t=A2jabCTteMz3HhD0-0)

- [x] updated the Deploy Preview link at the top of this PR (or remove it if not applicable)
- [x] updated the Figma referense link at the top of this PR (or remove it if not applicable)
- [x] tagged `@avaya-dux/dux-design` if any visual changes have occurred

This PR completes Part 1 of the new Table Filter functionality that now uses a Drawer instead of a Sheet.

`@avaya-dux/dux-design`: Verify the new default behavior for the Table Filter Columns

- Clicking on the Filter button will invoke the Drawer that lists all columns, those visible should be checked.
- Checking / Unchecking columns will not take effect until the Apply button is pressed.
- The Apply button will only be enabled when the current selections actually change.
- After the Apply button is pressed the drawer will close and the appropriate columns will show or hide.
- Pressing the Cancel button will close the drawer while ignoring any changes done to the selections.
- "**not in scope**" Filter button states are still being decided and will be implemented in Part 2.
- "**not in scope**" Ability to provide custom Drawer or any other UI will be implemented in Part 2.

`@avaya-dux/dux-devs`: See notes for UX designers above to see new behavior test cases.

- We need to keep 2 different set of column arrays. One for the current state and another for the new state that will either be saved when the user clicks the Apply button or thrown away if the user clicks the Cancel button.
-  There is also another pair for arrays that only contain column ids that detects selection changes so that we can enable or disable the Apply button accordingly.
- I also updated the Table Filter tests since they now should verify we are using a Drawer component instead of a Sheet and we now require the user to click on Apply for changes to take place.
- "not in scope" The ability to provide a callback for the developer to provide their own UI will be done in Part 2.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new column filtering option in the table component, enhancing interactivity.
  - Implemented a `Drawer` for filter options, improving user experience.
  - Added new buttons for "Apply" and "Cancel" actions in the filter interface.

- **Bug Fixes**
  - Updated testing logic to align with the new visibility and interaction states of the dialog in the table component.

- **Documentation**
  - Revised translations for filter-related toolbar actions for better clarity.

### DIFF
--- a/src/components/Table/Table.test.jsx
+++ b/src/components/Table/Table.test.jsx
@@ -729,12 +729,7 @@ describe("Table", () => {
 			);
 			expect(firstColumnSortButton).toBeVisible();
 
-			expect(getByRole("dialog")).not.toHaveClass(
-				"sheet-horizontal-slide-in-shim",
-			);
-			expect(getByRole("dialog")).not.toHaveClass(
-				"sheet-horizontal-slide-out-shim",
-			);
+			expect(getByRole("dialog")).not.toHaveClass("neo-drawer neo-drawer--isOpen");
 
 			await user.click(firstColumnSortButton);
 
@@ -742,16 +737,16 @@ describe("Table", () => {
 			expect(menuItems).toHaveLength(4);
 			await user.click(queryAllByRole("menuitem")[3]);
 
-			expect(getByRole("dialog")).toHaveClass("sheet-horizontal-slide-in-shim");
-			expect(getByRole("dialog")).not.toHaveClass(
-				"sheet-horizontal-slide-out-shim",
-			);
+			expect(getByRole("dialog")).toHaveClass("neo-drawer neo-drawer--isOpen");
 
 			const nameCheckbox = getByLabelText(FilledFields.columns[0].Header);
 			expect(nameCheckbox).toBeChecked();
 
 			await user.click(nameCheckbox);
 			expect(nameCheckbox).not.toBeChecked();
+
+			const applyButton = getByLabelText("Apply");
+			await user.click(applyButton);
 			expect(firstColumnSortButton).not.toBeVisible();
 		});
 
@@ -770,25 +765,20 @@ describe("Table", () => {
 				`button[aria-label="${FilledFields.translations.toolbar.filterColumns}"]`,
 			);
 
-			expect(getByRole("dialog")).not.toHaveClass(
-				"sheet-horizontal-slide-in-shim",
-			);
-			expect(getByRole("dialog")).not.toHaveClass(
-				"sheet-horizontal-slide-out-shim",
-			);
+			expect(getByRole("dialog")).not.toHaveClass("neo-drawer neo-drawer--isOpen");
 
 			await user.click(columnFilterButton);
 
-			expect(getByRole("dialog")).toHaveClass("sheet-horizontal-slide-in-shim");
-			expect(getByRole("dialog")).not.toHaveClass(
-				"sheet-horizontal-slide-out-shim",
-			);
+			expect(getByRole("dialog")).toHaveClass("neo-drawer neo-drawer--isOpen");
 
 			const nameCheckbox = getByLabelText(FilledFields.columns[0].Header);
 			expect(nameCheckbox).toBeChecked();
 
 			await user.click(nameCheckbox);
 			expect(nameCheckbox).not.toBeChecked();
+
+			const applyButton = getByLabelText("Apply");
+			await user.click(applyButton);
 			expect(firstColumnSortButton).not.toBeVisible();
 		});
 	});
--- a/src/components/Table/Table.test.jsx
+++ b/src/components/Table/Table.test.jsx
@@ -747,7 +747,7 @@ describe("Table", () => {
 			await user.click(nameCheckbox);
 			expect(nameCheckbox).not.toBeChecked();
 
-			const applyButton = getByLabelText("Apply");
+			const applyButton = getByLabelText(FilledFields.translations.toolbar.apply);
 			await user.click(applyButton);
 			expect(firstColumnSortButton).not.toBeVisible();
 		});
@@ -781,7 +781,7 @@ describe("Table", () => {
 			await user.click(nameCheckbox);
 			expect(nameCheckbox).not.toBeChecked();
 
-			const applyButton = getByLabelText("Apply");
+			const applyButton = getByLabelText(FilledFields.translations.toolbar.apply);
 			await user.click(applyButton);
 			expect(firstColumnSortButton).not.toBeVisible();
 		});
--- a/src/components/Table/Table.test.jsx
+++ b/src/components/Table/Table.test.jsx
@@ -729,7 +729,9 @@ describe("Table", () => {
 			);
 			expect(firstColumnSortButton).toBeVisible();
 
-			expect(getByRole("dialog")).not.toHaveClass("neo-drawer neo-drawer--isOpen");
+			expect(getByRole("dialog")).not.toHaveClass(
+				"neo-drawer neo-drawer--isOpen",
+			);
 
 			await user.click(firstColumnSortButton);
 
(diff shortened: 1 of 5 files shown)


src/components/Table/TableComponents/TableFilter.tsx
```diff
@@ -34,16 +42,100 @@ export const TableFilter = <T extends Record<string, any>>({
...
+	);
+
+	const handleColumnVisibilityChange = useCallback(
+		(e: React.ChangeEvent<HTMLInputElement>) => {
+			const { checked, value } = e.target;
+			const visibleColumnIds: IdType<T>[] = [];
+
+			//Update filteredColumns array checked value
+			const nextColumns = filteredColumns.map((col) => {
+				if (
```
enrique-prado, line 71: this special logic is for detecting changes to the visible column checkboxes. It's inside this loop to avoid having to write a separate function that does another redundant loop. This is basically a tradeoff between performance and code readability. This loop needs to run every time a user checks or unchecks a column so I wanted to keep it performant and not write a separate (more readable) function that would require to walk the entire array again.

src/components/Table/Table.test.jsx
```diff
@@ -742,16 +739,16 @@ describe("Table", () => {
...
-			);
+			expect(getByRole("dialog")).toHaveClass("neo-drawer neo-drawer--isOpen");
 
 			const nameCheckbox = getByLabelText(FilledFields.columns[0].Header);
 			expect(nameCheckbox).toBeChecked();
 
 			await user.click(nameCheckbox);
 			expect(nameCheckbox).not.toBeChecked();
+
+			const applyButton = getByLabelText("Apply");
```
joe-s-avaya, line 750: Don't use static strings as these are fragile and tests will be broken if the translation data is updated (which has happened numerous times as Matt+Justin+David ask for translation updates frequently). Instead, use `FilledFields.translations.*`. Ie:
```suggestion
			const applyButton = getByLabelText(FilledFields.translations.toolbar.apply);
```

joe-s-avaya: the same on line 784 (`const applyButton = getByLabelText("Apply");`).

src/components/Table/TableComponents/TableFilter.tsx
```diff
@@ -15,15 +17,21 @@ type TableFilterProps<T extends Record<string, any>> = {
...
 	translations,
 	instance,
 }: TableFilterProps<T>) => {
 	// translations
-	const clear = translations.clear || defaultTranslations.toolbar.clear;
-	const close =
-		translations.close || defaultTranslations.toolbar.close || "Close";
+	const apply = translations.apply || defaultTranslations.toolbar.apply;
+	const cancel =
+		translations.cancel || defaultTranslations.toolbar.cancel || "Cancel";
```
joe-s-avaya, line 34: Remove secondary `||` as it is redundant.
```suggestion
		translations.cancel || defaultTranslations.toolbar.cancel;
```

joe-s-avaya: I know that it's not your code (not really) but we must have missed this in previous PRs, and you're touching the code, so we should improve as we go. The whole [Tidy First](https://learning.oreilly.com/library/view/tidy-first/9781098151232/) ideology.

joe-s-avaya:
BUG: should not allow user to hide all columns

src/components/Table/TableComponents/TableFilter.tsx
```diff
@@ -15,15 +17,21 @@ type TableFilterProps<T extends Record<string, any>> = {
 	instance: TableInstance<T>;
 };
 
+type FilterColumns = {
+	id: string;
+	hdr: string | undefined | null;
```
joe-s-avaya, line 22: Is `hdr` header text? Either way, I request that you rename as I am still confused about what exactly it is after reading your updates.

joe-s-avaya: Wait, is it the checkbox display text? If so, `hdr` is misleading. Recommend renaming to `columnHeaderText`/`checkboxDisplayText`.

joe-s-avaya (approved):
There are a few minor issues, but nothing I want to block on. Looks pretty good. 

I would like to see you add tests for the new functionality. Even just a single test that asserts that you cannot "apply" unless real changes have been made.

Justin-Avaya (approved):

src/components/Table/TableComponents/TableFilter.tsx
```diff
@@ -34,16 +42,100 @@ export const TableFilter = <T extends Record<string, any>>({
...
-			style={{ color: "black" }}
-			variant="tertiary"
+	// filteredColumns is a temporary array that is used while the filter Drawer is open.
+	const [filteredColumns, setNewAllColumns] = useState<FilterColumns[]>([]);
+	const [originalVisibleColIds, setOriginalVisibleColIds] = useState<
+		IdType<T>[]
+	>([]);
+	const [applyBtnEnabled, setApplyBtnEnabled] = useState<boolean>(false);
+
+	const didColumnsSelectionsChange = useCallback(
```
yangAtSpoken, line 52: Columns could be deselected or selected.  This function does not deal with deselection. It handles only newly selected.

src/components/Table/TableComponents/TableFilter.tsx
yangAtSpoken: I do not think you need to go through all columns.  This event handler is called when a column in toggled.  The column id and on/off state are already passed in. You need to keep this information and that is all.

src/components/Table/TableComponents/TableFilter.tsx
```diff
@@ -58,43 +150,27 @@ export const TableFilter = <T extends Record<string, any>>({
...
 			/>
 
-			<Sheet
-				actions={buttons}
-				className="neo-table__filters--sheet"
-				open={filterSheetVisible}
-				slide={filterSheetVisible}
+			<Drawer
 				title={filterColumns}
+				open={filterSheetVisible}
```
yangAtSpoken, line 155: filterSheetVisible could be renamed.

yangAtSpoken (commented):
left comments

matthewAvaya:
- Move 'Apply' button on the drawer to be second after 'Cancel'
- Possibly a bug. There is a gap in the bottom part of the drawer when resizing the height of the web browser. The bottom part of the drawer should always be connected to the base.

src/components/Table/TableComponents/TableFilter.tsx
enrique-prado: yes but that doesn't work for for the case were you make changes and then "undo" your changes in which case the Apply button should be disabled. The "undo" case is what makes it tricky and forces you to compare against the original state of all checked columns.

src/components/Table/Table.test.jsx
enrique-prado: of course. fixed.

src/components/Table/TableComponents/TableFilter.tsx
enrique-prado: agreed. done.

src/components/Table/TableComponents/TableFilter.tsx
enrique-prado: fair enough. Renamed to checkboxDisplayText

src/components/Table/TableComponents/TableFilter.tsx
enrique-prado: You are right, it could be renamed to something better like filterDrawerOpen but unfortunately we cannot rename this because it is part of the IFilterContext interface and we have to maintain backwards compatibility.

enrique-prado:
> * Move 'Apply' button on the drawer to be second after 'Cancel'
> * Possibly a bug. There is a gap in the bottom part of the drawer when resizing the height of the web browser. The bottom part of the drawer should always be connected to the base.

@matthewAvaya Hi Matt, I have made the change to the "Apply" button.  I'm creating a new bug for the second issue.

matthewAvaya (approved):

PR #513, opened 2024-08-30
NEO-2175: Select closing bug fix

enrique-prado:
[Link to Select Searchable story](https://deploy-preview-513--neo-react-library-storybook.netlify.app/?path=/story/components-select--searchable)

- [x] updated the Deploy Preview link at the top of this PR (or remove it if not applicable)

This PR fixes a bug in both the Single Searchable & Multi Searchable select components that happens when a user clicks on it to open it and it closes itself.  The dropdown of options now remains open as expected.

- The bug was caused by the underlying onFocus implementation of downshift.
- Updated an existing test with a built in delay to make sure the select component remains open.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved focus handling in the MultiSelect and SingleSelect components to prevent unintended actions when interacting with dropdowns.
	- Added functionality to close the dropdown when the input field gains focus, enhancing user experience.

- **Tests**
	- Enhanced test reliability for the Select component by implementing asynchronous checks for the `aria-expanded` property, ensuring accurate behavior during interactions.

### DIFF
--- a/src/components/Select/InternalComponents/MultiSelectSearchable.tsx
+++ b/src/components/Select/InternalComponents/MultiSelectSearchable.tsx
@@ -43,10 +43,6 @@ export const MultiSelectSearchable = () => {
 		getToggleButtonProps();
 	const { id, onKeyDown, ...restInputProps } = getInputProps();
 
-	if (restInputProps.onFocus) {
-		restInputProps.onFocus = undefined; // Override downshift behavior to fix NEO-2175
-	}
-
 	// clear the search when dropdown closes (when the user selects an item or clicks away)
 	useEffect(() => {
 		if (isOpen === false) {
--- a/src/components/Select/InternalComponents/MultiSelectSearchable.tsx
+++ b/src/components/Select/InternalComponents/MultiSelectSearchable.tsx
@@ -43,6 +43,10 @@ export const MultiSelectSearchable = () => {
 		getToggleButtonProps();
 	const { id, onKeyDown, ...restInputProps } = getInputProps();
 
+	if (restInputProps.onFocus) {
+		restInputProps.onFocus = undefined; // Override downshift behavior to fix NEO-2175
+	}
+
 	// clear the search when dropdown closes (when the user selects an item or clicks away)
 	useEffect(() => {
 		if (isOpen === false) {
--- a/src/components/Select/Searchable.test.jsx
+++ b/src/components/Select/Searchable.test.jsx
@@ -58,9 +58,8 @@ describe("Select", () => {
 			const searchableElement = getByRole("combobox");
 			expect(searchableElement).toHaveAttribute("aria-expanded", "false");
 			fireEvent.click(toggleButton);
-			await vi.waitUntil(
-				() =>
-					expect(searchableElement).toHaveAttribute("aria-expanded", "true")
+			await vi.waitUntil(() =>
+				expect(searchableElement).toHaveAttribute("aria-expanded", "true"),
 			);
 			fireEvent.click(toggleButton);
 			expect(searchableElement).toHaveAttribute("aria-expanded", "false");
--- a/src/components/Select/Searchable.test.jsx
+++ b/src/components/Select/Searchable.test.jsx
@@ -1,4 +1,10 @@
-import { act, fireEvent, render, screen } from "@testing-library/react";
+import {
+	act,
+	fireEvent,
+	render,
+	screen,
+} from "@testing-library/react";
+import { vi } from "vitest";
 import userEvent from "@testing-library/user-event";
 import { axe } from "jest-axe";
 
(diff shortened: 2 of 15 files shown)


yangAtSpoken (changes requested):
Both searchable selects [here](https://neo-react-library-storybook.netlify.app/?path=/story/components-select--searchable) used to open on getting keyboard focus. 
 Now, this behavior is gone.  Is it possible to get it back?

src/components/Select/InternalComponents/MultiSelectSearchable.tsx
```diff
@@ -43,6 +43,10 @@ export const MultiSelectSearchable = () => {
 		getToggleButtonProps();
 	const { id, onKeyDown, ...restInputProps } = getInputProps();
 
+	if (restInputProps.onFocus) {
+		restInputProps.onFocus = undefined; // Override downshift behavior to fix NEO-2175
+	}
```
joe-s-avaya, line 48: nit: prefer one liner:
```suggestion
	delete restInputProps.onFocus; // remove downshift behavior to fix NEO-2175
```

src/components/Select/Searchable.test.jsx
```diff
@@ -49,15 +50,19 @@ describe("Select", () => {
...
 				"span.neo-multiselect-combo__header",
 			);
 			const searchableElement = getByRole("combobox");
 			expect(searchableElement).toHaveAttribute("aria-expanded", "false");
 			fireEvent.click(toggleButton);
-			expect(searchableElement).toHaveAttribute("aria-expanded", "true");
+			await vi.waitUntil(
+				() =>
+					expect(searchableElement).toHaveAttribute("aria-expanded", "true"),
+				{ timeout: 500 },
```
joe-s-avaya, line 64: you shouldn't need a timeout here, please remove and test

src/components/Select/InternalComponents/MultiSelectSearchable.tsx
joe-s-avaya: same in `SingleSelectSearchable.tsx`

enrique-prado: yes, that was my original implementation but biome flagged using delete as a performance issue and suggested I do the assignment to undefined. Thoughts?

yangAtSpoken: [This codesandbox](https://codesandbox.io/p/sandbox/usecombobox-action-props-z6qkg?file=%2Fsrc%2Findex.js)  shows how to override onFocus().

src/components/Select/Searchable.test.jsx
enrique-prado: removed timeout. It passes.

yangAtSpoken:
Not sure if you have seen [this](https://github.com/downshift-js/downshift/issues/1095), might be helpful.

src/components/Select/InternalComponents/MultiSelectSearchable.tsx
enrique-prado: @yangAtSpoken @joe-s-avaya  Yes Yang, that would be another way of doing it.  In our case we don't want to pass a new implementation but simply delete the one from downshift.  Using the codesandbox pattern my code would look something like this:
onFocus: () => undefined
or 
onFocus: () => {}

Any thoughts?

yangAtSpoken: Does it help if you call openMenu() in onFocus?  I would like to keep the current behavior when it gets the keyboard focus.

joe-s-avaya (approved):
there are a few things left to do, but you're aware of them so I'm approving ahead of your updates so that I don't block you

src/components/Select/InternalComponents/MultiSelectSearchable.tsx
enrique-prado: > the latter (`onFocus: () => {}`)

using the empty object causes a warning so I'll leave it as undefined.

enrique-prado: > Does it help if you call openMenu() in onFocus? I would like to keep the current behavior when it gets the keyboard focus.

I tried the openMenu() solution but it does not fix the bug. So far overriding with undefined seems to be the best solution.

enrique-prado:
> Not sure if you have seen [this](https://github.com/downshift-js/downshift/issues/1095), might be helpful.

I tried the openMenu trick but it didn't work.

src/components/Select/InternalComponents/MultiSelectSearchable.tsx
```diff
@@ -43,6 +43,10 @@ export const MultiSelectSearchable = () => {
 		getToggleButtonProps();
 	const { id, onKeyDown, ...restInputProps } = getInputProps();
```
yangAtSpoken, line 44:
```suggestion
	const { id, onKeyDown, ...restInputProps } = getInputProps({ onFocus: undefined });
```
This works for me on my local. It preserves the keyboard behavior.  Please verify.

yangAtSpoken: Unfortunately, this can not be reproduced.

yangAtSpoken (approved):

PR #529, opened 2024-09-10
DEV ONLY // quick cleanup to Pagination template story

joe-s-avaya:
Update the Pagination template story so that it handles when the records per page is increased. 

It's the same idea as the [SSR Table Pagination PR](https://github.com/avaya-dux/neo-react-library/pull/527). 

## Summary by CodeRabbit

- **New Features**
	- Enhanced pagination functionality to automatically adjust the displayed page when the number of items per page is increased, improving user navigation experience.

### DIFF
--- a/src/components/Pagination/Pagination.stories.tsx
+++ b/src/components/Pagination/Pagination.stories.tsx
@@ -47,17 +47,21 @@ const Template: Story = {
 			<main>
 				<h1>{header}</h1>
 
-				<section>
-					<Pagination
-						{...rest}
-						itemsPerPage={itemsPerPage}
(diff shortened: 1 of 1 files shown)


yangAtSpoken (approved):
Definitely more work for the user now.  But I am not sure if there is a better way.

joe-s-avaya:
> Definitely more work for the user now. But I am not sure if there is a better way.

It's the same amount of work as before, the Story was just broken. But yes to your point. This is an annoying thing that the user shouldn't have to deal with IMO, but I couldn't figure out a better way in the short time I had to think about this. 😞

